PR #44, opened 2019-08-21
Support module values in global values.yaml file

### LINKED ISSUE #37: Update documentation about modules enablement mechanism
We've forgotten to update the documentation in the #33 and now we have misleading information in LIFECYCLE.md (and maybe in other places as well). We should fix docs ASAP.

### LINKED ISSUE #38: Add support for upper-level `values.yaml` (common for all modules) in module enablement mechanism
1. We have different behavior in two mechanisms, but it is counterintuitive and will definitely mislead all the users. When we merge values the sequence is: `modules/values.yaml`, `modules/000-some-module-name/values.yaml`, ConfigMap. And for module enablement mechanism we are missing the first part (`modules/values.yaml`) for now.
2. Ability to choose whether the module is enabled or not in common `values.yaml` will allow implementing the scenario with a few sets of modules, enabled by default – for example, it will allow having "minimal" and "standard" sets of modules. Two do that will have to create two different `values.yaml` and put the right one in the well-known location during startup.

diafour:
- change documentation about `Enabled` keys
- huge refactoring of tests
- mocks for helm and kubeclient in common packages to use from different tests

Fixes #38 
Fixes #37


LIFECYCLE.md
```diff
@@ -40,21 +40,36 @@ Addon-operator makes a list of all enabled modules for their execution and a lis
 - during the start of addon-operator
 - when an event to restart all modules occurs (see [VALUES](VALUES.md)).
 
-For each module in the values.yaml and in the ConfigMap/addon-operator there is a key with the name of a module that may contain:
+Modules are disabled by default. Module can be enabled by a key with the name of a module suffixed by `Enabled`. This key should contains a boolean value and can be specified in these sources:
```
distol:
```suggestion
Modules are disabled by default. The module can be enabled by a key with the module name suffixed by `Enabled`. This key should contain a boolean value and can be specified in these sources:
```

LIFECYCLE.md
```diff
@@ -40,21 +40,36 @@ Addon-operator makes a list of all enabled modules for their execution and a lis
...
 
-- false
-- true or the object with parameters
+- $MODULES_DIR/values.yaml
+- `values.yaml` files in modules directories
+- ConfigMap/addon-operator
 
-The key may be absent. In this case, the `true` value is considered.
+Boolean values from values.yaml files and ConfigMap/addon-operator are combined and if result is equal to `false` or is empty, then the module is disabled.
```
distol:
```suggestion
Boolean values from values.yaml files and ConfigMap/addon-operator are combined and if the result is equal to `false` or is empty, then the module is disabled.
```

LIFECYCLE.md
```diff
@@ -40,21 +40,36 @@ Addon-operator makes a list of all enabled modules for their execution and a lis
...
 
-If combined values from values.yaml and ConfigMap/addon-operator equal to `false`, then the module is disabled.
-
-If the value is not `false`, the additional check is conducted – the `enabled` script is executed (see below). If there is a script and it returns `false`, then the module is considered disabled.
+If the value is `true`, the additional check is performed – the `enabled` script is executed (see below). If script is present in module and it returns `false`, then the module is considered disabled. If script is not present or returns `true`, then module is enabled.
```
distol:
```suggestion
If the value is `true`, the additional check is performed – the `enabled` script is executed (see below). If the script is present in the module and it returns `false`, then the module is considered disabled. If the script is not present or returns `true`, then the module is enabled.
```

LIFECYCLE.md
```diff
@@ -40,21 +40,36 @@ Addon-operator makes a list of all enabled modules for their execution and a lis
...
 
 If an error occurs during the modules discovery process, then the module discovery is restarted every 5 seconds until successful execution. In this case, the execution of hooks with `schedule` and `onKubernetesEvent` bindings will be blocked.
 
 As a result of a module discovery process the tasks for the execution of all *enabled* modules, deletion of all *disabled* modules, and execution of all global hook with the `afterAll` binding are added to the queue.
 
+### Module enable examples
```
distol:
```suggestion
### Module enablement examples
```

LIFECYCLE.md
```diff
@@ -75,6 +90,42 @@ fi
...
+  someModuleEnabled: "true"
+
+$ cat modules/01-some-module/enabled
+
+#!/bin/bash
+
+echo false > $MODULE_ENABLED_RESULT
+```
+
+Module some-module is explicitly disabled in modules/values.yaml and enabled by `someModuleEnabled` key in ConfigMap/addon-operator. So enabled script is executed. Script returns `false` and the final result is that module is disabled.
```
distol:
```suggestion
Module some-module is explicitly disabled in modules/values.yaml and enabled by `someModuleEnabled` key in ConfigMap/addon-operator. So enabled script is executed. Script returns `false` and the final result is that the module is disabled.
```

VALUES.md
```diff
@@ -73,9 +74,9 @@ Patch for temporary updates is returned via $VALUES_JSON_PATCH_PATH file and rem
 
 # Merged values
 
-A merged set of values is generated by Addon-operator when hooks or `enabled` scripts are executed or when helm charts are installed. In order to do that the global values are combined with the module values from the values.yaml and ConfigMap/addon-operator, and then the patches for the temporary update are applied.
+When hook or `enabled` script is executed or helm chart is installed, Addon-operator generates a merged set of values. This merged set combines global values from values.yaml files and ConfigMap/addon-operator and module values from the values.yaml files and ConfigMap/addon-operator. Also patches for the temporary update are applied.
```
distol: When the hook or `enabled` script should be executed, or helm chart is going to be installed, Addon-operator generates a merged set of values. This merged set combines:
* global values from values.yaml files and ConfigMap/addon-operator;
* module values from the values.yaml files and ConfigMap/addon-operator;
* patches for the temporary update are applied.

VALUES.md
```diff
@@ -73,9 +74,9 @@ Patch for temporary updates is returned via $VALUES_JSON_PATCH_PATH file and rem
...
 
-Next, the merged values are passed as the temporary JSON file to hooks or `enabled` script and as a temporary values.yaml file to the helm chart.
+The merged values are passed as the temporary JSON file to hooks or `enabled` script and as a temporary values.yaml file to the helm chart.
```
distol:
```suggestion
The merged values are passed as the temporary JSON file to hooks or `enabled` script and as the temporary values.yaml file to the helm installing the chart.
```

PR #50, opened 2019-09-24
feat: use werf to install modules' charts

diafour:


diafour:
Its time to close this first attempt to use werf with addon-operator.

PR #69, opened 2019-12-19
feat: include snapshots

diafour:


diafour:
Merged in combined commit #72

PR #71, opened 2020-01-20
feat: reload module run on values changes

diafour:


diafour:
Merged in combined commit #72

Issue #88, opened 2020-02-14
Improve logging fields


diafour:
1. global 'kubernetes' hook can trigger 'modules reload' and ModuleRun tasks will have 'hook' field with the name of hook that triggered an event but BeforeAll and AfterAll tasks will have 'hook' field with that hook name. An additional field is required, for example, 'event.triggered-by'.

2. 'module' field is empty for ModuleHookRun tasks and for 'queue task ModuleHookRun' records.

3. No information about an event for module 'kubernetes' hooks. It will be helpful if log record contain fields with binding name and watchEvent.

diafour:
1.
```
    {
      "binding": "beforeAll",
      "event.id": "dbe17ce6-210c-43a9-985d-36b845de97dc",
-->   "event.triggered-by.binding": "schedule",
-->   "event.triggered-by.binding.name": "schedule",
-->   "event.triggered-by.hook": "v1-hook-schedule.sh",
      "hook": "v1-hook-beforeall.sh",
      "hook.type": "global",
      "level": "info",
      "msg": "queue GlobalHookRun task",
      "queue": "main",
      "task.id": "735abd7e-c8ea-4dd7-953d-05ae28701dda",
      "time": "2020-02-20T09:54:11Z"
    }

```

3.
```
    {
      "binding": "kubernetes",
-->   "binding.name": "pods-all",
      "event.id": "2bb052de-8a99-4c7e-8983-fec3228bf9e0",
      "hook": "v1-hook-kube.sh",
      "hook.type": "global",
      "level": "info",
      "msg": "Run global hook",
      "operator.component": "taskRunner",
      "queue": "pods-all",
      "task.id": "9afbeae8-dd13-4cc6-ad1a-f8c087a4d8ad",
      "time": "2020-02-20T09:48:14Z",
-->   "watchEvent": "Modified"
   }

```

PR #91, opened 2020-02-19
fix: labels for json logs

diafour:
- add module label for logs of ModuleHookRun tasks
- add binding info for hook run tasks
- relabel when GlobalHookRun triggers a CreateReloadAll


diafour:
fixes #88

PR #114, opened 2020-04-03
fix: add --tiller-connection-timeout for `helm version` command

diafour:
- default timeout was 300s, set it to 2s.
- timeout of tiller waiting process is 10s.


diafour:
Fix for suddenly merged #113.

Issue #144, opened 2020-08-27
Error in resolving Helm version 


dnbstd:
## Hello Dev team!

**Original Dockerfile from repository with changed kubectl and helm version:**
```dockerfile
 # build libjq
FROM ubuntu:18.04 AS libjq
ENV DEBIAN_FRONTEND=noninteractive \
    DEBCONF_NONINTERACTIVE_SEEN=true \
    LC_ALL=C.UTF-8 \
    LANG=C.UTF-8

RUN apt-get update && \
    apt-get install -y git ca-certificates && \
    git clone https://github.com/flant/libjq-go /libjq-go && \
    cd /libjq-go && \
    git submodule update --init --recursive && \
    /libjq-go/scripts/install-libjq-dependencies-ubuntu.sh && \
    /libjq-go/scripts/build-libjq-static.sh /libjq-go /libjq


# build addon-operator binary linked with libjq
FROM golang:1.14 AS addon-operator
ARG appVersion=latest

# Cache-friendly download of go dependencies.
ADD go.mod go.sum /addon-operator/
WORKDIR /addon-operator
RUN go mod download

COPY --from=libjq /libjq /libjq
ADD . /addon-operator
WORKDIR /addon-operator

RUN git submodule update --init --recursive && ./go-build.sh $appVersion

FROM krallin/ubuntu-tini:bionic AS tini

# build final image
FROM ubuntu:18.04
RUN apt-get update && \
    apt-get install -y ca-certificates wget jq && \
    rm -rf /var/lib/apt/lists && \
    wget https://storage.googleapis.com/kubernetes-release/release/v1.17.9/bin/linux/amd64/kubectl -O /bin/kubectl && \
    chmod +x /bin/kubectl && \
    wget https://storage.googleapis.com/kubernetes-helm/helm-v2.16.10-linux-amd64.tar.gz -O /helm.tgz && \
    tar -z -x -C /bin -f /helm.tgz --strip-components=1 linux-amd64/helm linux-amd64/tiller && \
    rm -f /helm.tgz && \
    helm init --client-only && \
    mkdir /hooks
COPY --from=tini /usr/local/bin/tini /sbin/tini
COPY --from=addon-operator /addon-operator/addon-operator /
WORKDIR /
ENV MODULES_DIR /modules
ENV GLOBAL_HOOKS_DIR /global-hooks
```
**Addon-operator is recognize Helm version >2.16 as Helm 3**

part of log with debug on:
```log
[...]

PR #146, opened 2020-09-04
Fix detect helm version

dnbstd:
To fix bug with detect Helm3 version [(isssue)](https://github.com/flant/addon-operator/issues/144) this PR change function **InitAndVersion** in helm3.go.
If function recognize that version of Helm is 2 than function is return error ""detected Helm2 client version"", after that the addon operator is correctly initialized.


diafour:
Some conclusions from a conversation with @dnbstd :
- It seems that `helm version --template='{{ .Client.SemVer }}'` is not working for helm2 and helm3 has no `.Client.SemVer` field.
- A hotfix with little changes is not suited for this problem.

So, this PR is closed in favor of #147.

PR #161, opened 2020-12-14
fix: do not count the first task as a task from queue

nabokihms:
This MR fixes the race condition in ModuleRun tasks handling.

### DIFF
--- a/pkg/addon-operator/operator.go
+++ b/pkg/addon-operator/operator.go
@@ -620,7 +620,7 @@ func (op *AddonOperator) StartModuleManagerEventHandler() {
 					logEntry := eventLogEntry.WithFields(utils.LabelsToLogFields(logLabels))
 					for _, moduleChange := range moduleEvent.ModulesChanges {
 						// Do not add ModuleRun task if it is already queued.
-						hasTask := QueueHasModuleRunTask(op.TaskQueues.GetMain(), moduleChange.Name)
+						hasTask := QueueHasPendingModuleRunTask(op.TaskQueues.GetMain(), moduleChange.Name)
 						if !hasTask {
 							logEntry.WithField("module", moduleChange.Name).Infof("module values are changed, queue ModuleRun task")
 							newLabels := utils.MergeLabels(logLabels)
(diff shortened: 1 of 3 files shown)


pkg/addon-operator/operator.go
```diff
@@ -2058,14 +2058,26 @@ func InitAndStart(operator *AddonOperator) error {
 
 func QueueHasModuleRunTask(q *queue.TaskQueue, moduleName string) bool {
```
diafour, line 2059:
```suggestion
// QueueHasPendingModuleRunTask returns true if queue has pending tasks
// with the type "ModuleRun" related to the module "moduleName".
func QueueHasPendingModuleRunTask(q *queue.TaskQueue, moduleName string) bool {
```

pkg/addon-operator/operator.go
```diff
@@ -2058,14 +2058,26 @@ func InitAndStart(operator *AddonOperator) error {
...
-		if t.GetType() == task.ModuleRun {
-			hm := task.HookMetadataAccessor(t)
-			if hm.ModuleName == moduleName {
-				hasTask = true
-			}
+	firstTask := true
+
+	q.Iterate(func(t sh_task.Task) {
+		// Skip the first task in the queue.
+		// The first task is actually not in the queue anymore, because it has been already executed.
```
diafour, line 2065:
```suggestion
		// Skip the first task in the queue as it can be executed already, i.e. "not pending".
```

pkg/addon-operator/operator.go
```diff
@@ -2058,14 +2058,26 @@ func InitAndStart(operator *AddonOperator) error {
...
+		if t.GetType() != task.ModuleRun {
+			return
 		}
-		return true
+
+		if task.HookMetadataAccessor(t).ModuleName != moduleName {
+			return
+		}
+
+		hasTask = true
```
diafour, line 2079: Uh-oh, I don't like this "inverted logic" style ;) Is it possible to revert to a previous condition `if type == ModuleRun && name == moduleName { hasTask = true }`?

go.mod
```diff
@@ -5,7 +5,7 @@ go 1.12
 require (
 	github.com/davecgh/go-spew v1.1.1
 	github.com/evanphx/json-patch v4.9.0+incompatible
-	github.com/flant/shell-operator v1.0.0-beta.13.0.20201213124752-17d64f6e61d1 // branch: master
+	github.com/flant/shell-operator v1.0.0-beta.13.0.20201214165802-7644ce7cef6e // branch: master
```
diafour, line 8:
```suggestion
	github.com/flant/shell-operator v1.0.0-beta.13.0.20201214204115-463c4ecd8214 // branch: master
```

More recent master to prevent merge conflict with https://github.com/flant/addon-operator/pull/160

PR #168, opened 2021-01-15
fix: remove defaulting from config values

diafour:
#### Overview

Revert defaulting for config values.

`$CONFIG_VALUES_PATH` file should contain only ConfigMap content to hook be able to create proper config patches.

No

```release-note

```

### DIFF
--- a/VALUES.md
+++ b/VALUES.md
@@ -227,13 +227,15 @@ would generate the string `replicas: 200`. As you can see, the value "100" from 
 
 # Validation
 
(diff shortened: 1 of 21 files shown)


diafour:
Turns out that support for default is not enough.

We need a contract for Helm values, but values should be validated after each hook execution. So, we plan to implement another x- field: `x-required-for-helm`. Values are validated for `type` and `additionalFields` after hook execution and values are validated with required before helm execution.
This feature is a lifesaver when Helm values are constructed by multiple hooks.

Another enhancement — use `additionalFields: false` by default for schemas with `type: object` to save some time.

PR #173, opened 2021-02-10
feat(docs): add schemas with values flow

diafour:
#### Overview

Add schemas with values flow.

These schemas should help to better understand values flow.

```release-note

```


nabokihms (approved):
Great work!

PR #176, opened 2021-03-02
feat: execution rate settings for hooks

diafour:
#### Overview

Bring in execution rate settings from shell-operator: https://github.com/flant/shell-operator/pull/257.

Add `HELM_IGNORE_RELEASE` option to not treat the addon-operator release as a module.

1. Addon-operator can combine tasks in the queue and run the hook once with an array of binding contexts. Execution rate settings help to handle multiple events that are occurred during a short period.

2. `HELM_IGNORE_RELEASE` prevents self-destcruction in setup where addon-operator release is stored in the same namespace as modules (`ADDON_OPERATOR_NAMESPACE`).

```release-note

```

### DIFF
--- a/HOOKS.md
+++ b/HOOKS.md
@@ -38,12 +38,11 @@ During execution, a module hook receives global values and module values. Module
 
 ### onStartup
 
-Example JSON syntax:
+Example:
 
-```json
-{
-  "onStartup": ORDER
-}
+```yaml
+configVersion: v1
+onStartup: ORDER
 ```
 
 Parameters:
@@ -52,12 +51,11 @@ Parameters:
 
 ### beforeAll
 
-Example JSON syntax:
+Example:
 
-```json
-{
-  "beforeAll": ORDER
-}
+```yaml
+configVersion: v1
+beforeAll: ORDER
 ```
 
 Parameters:
@@ -66,12 +64,11 @@ Parameters:
 
 ### afterAll
 
-Example JSON syntax:
+Example:
 
-```json
-{
-  "afterAll": ORDER
-}
+```yaml
+configVersion: v1
+afterAll: ORDER
 ```
 
 Parameters:
@@ -80,12 +77,11 @@ Parameters:
 
 ### beforeHelm
 
-Example JSON syntax:
+Example:
 
-```json
-{
-  "beforeHelm": ORDER
-}
+```yaml
+configVersion: v1
+beforeHelm: ORDER
 ```
 
 Parameters:
@@ -94,12 +90,11 @@ Parameters:
 
 ### afterHelm
 
-Example JSON syntax:
+Example:
 
-```json
-{
-  "afterHelm": ORDER
-}
+```yaml
+configVersion: v1
+afterHelm: ORDER
 ```
 
 Parameters:
@@ -108,12 +103,11 @@ Parameters:
 
 ### afterDeleteHelm
 
-Example JSON syntax:
+Example:
 
-```json
-{
-  "afterDeleteHelm": ORDER
-}
+```yaml
+configVersion: v1
+afterDeleteHelm: ORDER
 ```
 
 Parameters:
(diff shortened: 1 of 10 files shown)


HOOKS.md
```diff
@@ -165,7 +159,7 @@ kubernetes:
 
 This hook will be executed *before* updating the Helm release with this binding context:
 
-```json
```
nabokihms, line 168: it looks like we need to return `json` back here.

pkg/addon-operator/operator.go
```diff
@@ -1256,6 +1256,16 @@ func (op *AddonOperator) HandleModuleHookRun(t sh_task.Task, labels map[string]s
 	hm := task.HookMetadataAccessor(t)
 	taskHook := op.ModuleManager.GetModuleHook(hm.HookName)
 
+	err := taskHook.GetRateLimiter().Wait(context.Background())
```
nabokihms, line 1259: Maybe we can hide the getter method `GetRateLimiter()` to forbid other interactions with the rate limiter.

Before:
```go
func (t *taskHook) GetRateLimiter() *rate.Limiter {
  return t.RateLimiter
}
```

After:
```go
func (t *taskHook) RateLimitWait() err {
  return t.rateLimiter.Wait(context.Background())
}
```

RUNNING.md
```diff
@@ -91,6 +91,14 @@ Addon-operator expects that "helm" binary is available in $PATH. It detects Helm
 
 **HELM3** — set to "yes" to disable auto-detection and explicitly enable compatibility with helm3.
 
+**HELM_IGNORE_RELEASE** — a name of the release that should not be treated as the module's release. Prevent self-destruction when addon-operator release is stored in the same namespace as releases for modules.
```
nabokihms, line 94:
```suggestion
**HELM3_IGNORE_RELEASE** — a name of the release that should not be treated as the module's release. Prevent self-destruction when addon-operator release is stored in the same namespace as releases for modules.
```
?

diafour: Yeah, it is implemented only for Helm3 now, but I think it will affect future versions. I wish that we don't need that "ignored release" at all. There should be a mechanism to filter only "our" releases with labels or prefixes.

Helm 2 setup is affected if the cluster-wide tiller's namespace is equal to ADDON_OPERATOR_NAMESPACE. No one stops you from deploying an addon-operator to the same namespace as the tiller. Also, there are situations with "external" tiller, e.g. `werf`.

pkg/addon-operator/operator.go
diafour: The context should be exposed, so it's more like that:

```go
func (t *taskHook) RateLimitWait(ctx context) err {
  return t.rateLimiter.Wait(ctx)
}
```

`RateLimiter` is public to use it from shell-operator and from addon-operator.

HOOKS.md
diafour: It's a hack to ignore json's whining about "...".

diafour: I'll put more spaces and it will look good.

nabokihms: I get it.

RUNNING.md
nabokihms: We decided not to rename it.

pkg/app/app.go
```diff
@@ -89,6 +90,10 @@ func DefineStartCommandFlags(kpApp *kingpin.Application, cmd *kingpin.CmdClause)
 		Default(Helm3Timeout.String()).
 		DurationVar(&Helm3Timeout)
 
+	cmd.Flag("helm-ignore-release", "Helm3+: do not count release as a module release.").
```
zuzzas, line 93: This change should be mentioned in the description of the PR.

diafour: Done

nabokihms (approved):

PR #177, opened 2021-03-02
Kubernetes object manipulation

zuzzas:

```release-note

```

### DIFF
--- a/pkg/addon-operator/operator.go
+++ b/pkg/addon-operator/operator.go
@@ -12,10 +12,11 @@ import (
 	"strings"
 	"time"
 
+	"github.com/flant/shell-operator/pkg/kube/object_patch"
 	"github.com/go-chi/chi"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	log "github.com/sirupsen/logrus"
-	"gopkg.in/satori/go.uuid.v1"
+	uuid "gopkg.in/satori/go.uuid.v1"
 	"sigs.k8s.io/yaml"
 
 	sh_app "github.com/flant/shell-operator/pkg/app"
@@ -24,7 +25,7 @@ import (
 	. "github.com/flant/shell-operator/pkg/hook/types"
 	"github.com/flant/shell-operator/pkg/kube_events_manager/types"
 	"github.com/flant/shell-operator/pkg/metric_storage"
-	"github.com/flant/shell-operator/pkg/shell-operator"
+	shell_operator "github.com/flant/shell-operator/pkg/shell-operator"
 	sh_task "github.com/flant/shell-operator/pkg/task"
 	"github.com/flant/shell-operator/pkg/task/queue"
 	. "github.com/flant/shell-operator/pkg/utils/measure"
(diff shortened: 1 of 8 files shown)


pkg/module_manager/module_hook.go
```diff
@@ -336,3 +355,12 @@ func (h *ModuleHook) prepareMetricsFile() (string, error) {
 	}
 	return path, nil
 }
+
+// KUBERNETES PATCH PAH
```
diafour, line 359:
```suggestion
// KUBERNETES PATCH PATH
```

pkg/module_manager/module_hook.go
```diff
@@ -165,6 +166,19 @@ func (h *ModuleHook) Run(bindingType BindingType, context []BindingContext, logL
...
+		specs, err = h.moduleManager.KubeObjectPatcher.ParseSpecs(hookResult.KubernetesPatchBytes)
+		if err != nil {
+			return err
+		}
+
+		err = h.moduleManager.KubeObjectPatcher.GenerateFromJSONAndExecuteOperations(specs)
+		if err != nil {
+			return err
+		}
+	}
```
diafour, line 180: Can it be just one method, as a `SendBatch` for metrics?

zuzzas: Have I made the same mistake in the shell-operator PR?

diafour: Yes, the same logic applies to a `Hook.Run` method in the shell-operator.

https://github.com/flant/shell-operator/pull/239/files#diff-d6e8286d4f35ebf5243c1d71976c0114b0d7db3bf7dca16dbfb5f034b351ec4cR137-R156

zuzzas: I believe, there is a simpler solution. Also pushed it to the shell-operator PR.
https://github.com/flant/addon-operator/pull/177/commits/c8b67e5533f01705cce6cc243145b83c07e5cc59

diafour: Yeah, it is better now.

The problem that I still see here is that the `KubernetesPatchBytes` in hookResult affect go-hooks in a way that we need to generate and return patches in `[]byte` form instead of handy struct [OperationSpec](https://github.com/flant/shell-operator/pull/239/files#diff-726461c73ad5fd14bc1f28d156c4fc31e5675c340c767552491edd9b40f73fd4R3-R15). See https://github.com/flant/addon-operator/blob/master/sdk/sdk.go#L53-L58. _As I know, there is a new sdk, so maybe bytes in HookResult are not an issue._

pkg/module_manager/module_manager.go
```diff
@@ -9,6 +9,7 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/flant/shell-operator/pkg/kube/object_patch"
```
diafour, line 12: Unsorted import

diafour (changes requested):

pkg/addon-operator/operator.go
```diff
@@ -173,6 +174,7 @@ func (op *AddonOperator) InitModuleManager() error {
 	op.ModuleManager.WithKubeConfigManager(op.KubeConfigManager)
 	op.ModuleManager.WithScheduleManager(op.ScheduleManager)
 	op.ModuleManager.WithKubeEventManager(op.KubeEventsManager)
+	op.ModuleManager.WithKubeObjectPatcher(object_patch.NewObjectPatcher(op.KubeClient))
```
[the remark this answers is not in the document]
zuzzas: Done!

pkg/module_manager/module_hook.go
zuzzas: Ah, that makes sense. Fixed here and in the shell-operator!

diafour (approved):

PR #181, opened 2021-03-12
fix: fresh objects for Synchronization executions

diafour:
#### Overview

Set `objects` in "Synchronization" binding context to an actual state in the cluster.

See https://github.com/flant/shell-operator/pull/261

```release-note

```


pkg/addon-operator/operator.go
```diff
@@ -765,15 +766,19 @@ func (op *AddonOperator) TaskHandler(t sh_task.Task) queue.TaskResult {
...
+					ExecuteOnSynchronization: info.KubernetesBinding.ExecuteHookOnSynchronization,
 				})
+
+			// Ignore "waitForSynchronization: true" if Synchronization is not required.
 			if info.QueueName == t.GetQueueName() {
 				mainSyncTasks = append(mainSyncTasks, newTask)
 			} else {
-				if info.WaitForSynchronization {
-					waitSyncTasks[kubernetesBindingId] = newTask
+				if info.KubernetesBinding.WaitForSynchronization && info.KubernetesBinding.ExecuteHookOnSynchronization {
```
nabokihms, line 780: Oh, I thought that `waitForSynchronization` controls the moment when we start firing monitor events. It can be tricky that `executeHookOnSynchronization` also affects it.

pkg/addon-operator/operator.go
```diff
@@ -839,6 +841,12 @@ func (op *AddonOperator) TaskHandler(t sh_task.Task) queue.TaskResult {
 			op.ModuleManager.DumpState()
 			t.WithQueuedAt(time.Now())
 			res.Status = "Repeat"
+		} else {
+			taskLogEntry.Infof("Global hooks enable kubernetes events")
+			//op.ModuleManager.GlobalEnableKubernetesEvents()
```
nabokihms, line 846: I think we can remove commented lines here.

pkg/addon-operator/operator.go
```diff
@@ -1279,10 +1291,23 @@ func (op *AddonOperator) HandleModuleHookRun(t sh_task.Task, labels map[string]s
 		op.MetricStorage.HistogramObserve("{PREFIX}module_hook_run_seconds", d.Seconds(), metricLabels)
 	})()
 
-	logEntry.Infof("Module hook start '%s'", hm.HookName)
+	isSynchronization := hm.BindingType == OnKubernetesEvent && hm.BindingContext[0].Type == types.TypeSynchronization
```
nabokihms, line 1294: Could `hm.BindingContext[0].Type` be equal to Group?

pkg/addon-operator/operator.go
```diff
@@ -1326,46 +1362,56 @@ func (op *AddonOperator) HandleModuleHookRun(t sh_task.Task, labels map[string]s
...
-		defer op.HelmResourcesManager.ResumeMonitor(hm.ModuleName)
-	}
+	if shouldRunHook {
+		logEntry.Infof("Module hook start '%s'", hm.HookName)
 
-	err = op.ModuleManager.RunModuleHook(hm.HookName, hm.BindingType, hm.BindingContext, t.GetLogLabels())
+		// Module hook can recreate helm objects, so pause resources monitor.
+		// Parallel hooks can interfere, so pause-resume only for hooks in the main queue.
+		// FIXME pause-resume for parallel hooks
+		if t.GetQueueName() == "main" {
```
nabokihms, line 1371: It is not connected to the PR subject, but I just curious about setting `queue: main` in a hook config. Is it allowed? Don't we want to forbid explicitly setting queue to "main"?

pkg/addon-operator/operator.go
```diff
@@ -1398,10 +1444,23 @@ func (op *AddonOperator) HandleGlobalHookRun(t sh_task.Task, labels map[string]s
 		op.MetricStorage.HistogramObserve("{PREFIX}global_hook_run_seconds", d.Seconds(), metricLabels)
 	})()
 
-	logEntry.Infof("Global hook run")
+	isSynchronization := hm.BindingType == OnKubernetesEvent && hm.BindingContext[0].Type == types.TypeSynchronization
+	shouldRunHook := true
```
nabokihms, line 1448: Can we move this code to a method for this instead of copying it? And why do we add `shouldRunHook` by the way? :)

pkg/addon-operator/operator.go
diafour: `waitForSynchronization: true` is to postpone `beforeHelm` and `beforeAll` until "Synchronization" is done in parallel queues. It is not an obvious name, yes ;)

pkg/addon-operator/operator.go
diafour: No, BindingContext here is always event-specific. `"type":"Group"` is generated lately.

pkg/addon-operator/operator.go
diafour: A good rule for future linting. Also, we can use something different instead of the "main" string.

pkg/addon-operator/operator.go
diafour: 1. I've created a method IsSynchronization.

2. `shouldRunHook` is needed as GlobalHookRun and ModuleHookRun tasks are now created for bindings with `executeHookOnSynchronization: false`. These tasks only enable emitting events from the binding's Monitor and do not run the hook.

nabokihms: About the second part:
I was wondering about moving some code out of the `Handle*` methods. If it is hard or gives no benefits, let's leave things as is.

nabokihms (approved):
No serious concerns left. The only thing we have to do is updating `go.mod` after merging https://github.com/flant/shell-operator/pull/261

PR #182, opened 2021-03-12
feat: use separate kube client for helm resources monitor

diafour:
#### Overview

Add a separate kube client for Helm resources monitor.

The addon-operator uses the same kube client for hook events and for Helm resources monitor. Helm resources monitor periodically lists all resources defined in Helm charts and kube client becomes rate limited. This PR adds new flags to set separate rate limit settings for Helm resources monitor.

```release-note

```

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -26,6 +26,10 @@ var TillerMaxHistory int32 = 0
 var Helm3HistoryMax int32 = 10
 var Helm3Timeout time.Duration = 5 * time.Minute
 var HelmIgnoreRelease = ""
+var HelmMonitorKubeClientQpsDefault = "5" // DefaultQPS from k8s.io/client-go/rest/config.go
+var HelmMonitorKubeClientQps float32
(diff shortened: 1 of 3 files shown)


pkg/app/app.go
```diff
@@ -94,6 +98,16 @@ func DefineStartCommandFlags(kpApp *kingpin.Application, cmd *kingpin.CmdClause)
 		Envar("HELM_IGNORE_RELEASE").
 		StringVar(&HelmIgnoreRelease)
 
+	// Rate limit settings for kube client used by Helm resources monitor.
```
nabokihms, line 101: Commentary is right, but I can figure out the same from the name of flags or env vars. The `HELM_KUBE_CLIENT_` is a little confusing prefix because it does not affect the helm, actually.

nabokihms (approved):
The only concern I have is about settings naming. Everything else looks good. Approved in advance.

diafour:
No new logic, only renames.

pkg/app/app.go
diafour: HELM_MONITOR_ do the thing.

PR #183, opened 2021-03-13
fix: ignore 'nonexistent path' errors from 'remove' patch operations

diafour:
#### Overview

Ignore errors from 'remove' patch operations.

It is a common situation when config values contain a key, then hook returns a patch and removes this key. Lately, user edits a ConfigMap and also removes a key. Since then, the patch set is "poisoned" with the 'remove' operation of a nonexistent key.

```release-note

```

### DIFF
--- a/pkg/utils/values_patch.go
+++ b/pkg/utils/values_patch.go
@@ -28,6 +28,7 @@ func NewValuesPatch() *ValuesPatch {
 	}
 }
 
+// ToJsonPatch returns a jsonpatch.Patch with all operations.
 func (p *ValuesPatch) ToJsonPatch() (jsonpatch.Patch, error) {
 	data, err := json.Marshal(p.Operations)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


pkg/utils/values_patch.go
```diff
@@ -243,16 +285,25 @@ func CompactPatches(operations []*ValuesPatchOperation) ValuesPatch {
 	return newValuesPatch
 }
 
+type ApplyPatchOptions string
+
+const Strict ApplyPatchOptions = "strict"
+
 // ApplyValuesPatch applies a set of json patch operations to the values and returns a result
-func ApplyValuesPatch(values Values, valuesPatch ValuesPatch) (Values, bool, error) {
+func ApplyValuesPatch(values Values, valuesPatch ValuesPatch, options ...ApplyPatchOptions) (Values, bool, error) {
```
nabokihms, line 293: Could we avoid optional arguments here?

pkg/utils/values_patch.go
```diff
@@ -40,15 +41,47 @@ func (p *ValuesPatch) ToJsonPatch() (jsonpatch.Patch, error) {
...
+	for _, op := range p.Operations {
+		patch, err := op.ToJsonPatch()
+		if err != nil {
+			return nil, err
+		}
+		newDoc, err := patch.Apply(doc)
+		// Ignore errors for remove operation.
+		if err != nil && op.Op == "remove" {
+			errStr := err.Error()
+			if strings.HasPrefix(errStr, "error in remove for path:") {
```
nabokihms, line 69: Is this something we can rely on? I assume we check prefix here only to ignore `"remove operation failed to decode path"` error from the `op.Path` method.

Maybe we can validate paths of operations separately from the `patch.Apply` method and continue on every error here. WDYT?

pkg/utils/values_patch_test.go
```diff
@@ -399,26 +565,46 @@ func Test_jsonpatch_Remove_ObjectAndArray(t *testing.T) {
...
+// A "replace" patch operation referencing a key that does not exist
+// in the target object are currently being accepted; ultimately becoming
+// an "add" operation on the target object. This is incorrect per the specification.
+//
+// 11.2020 the PR was merged but then reverted in https://github.com/evanphx/json-patch/pull/111
+//
+// The fix is in v5 now. So this test should work with v5.
 func Test_jsonpatch_Replace_NonExistent_IsError(t *testing.T) {
-	t.SkipNow() // Test is no working as expected
+	t.SkipNow()
```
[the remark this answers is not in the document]
diafour: A more clear explanation is added.

pkg/utils/values_patch.go
diafour: "options" arguments are replaced with the "mode" argument and two modes are defined: Strict and IgnoreNonExistentPaths.

pkg/utils/values_patch.go
diafour: As discussed, prefixes are extracted as constants, and error checking is extracted to the function.

nabokihms (approved):
Neat, thanks!

PR #200, opened 2021-05-05
Refactor gabs

yalosev:
#### Overview

- Remove gabs library
- Add gjson library
- Add methods to get values with gjson

Using one lib for runtime and tests
Introduce change only for go-hooks


zuzzas (approved):

pkg/module_manager/go_hook/patchable_values.go
```diff
@@ -1,25 +1,63 @@
...
+	v := p.values.Get(path)
+	if v.Exists() {
+		return v, true
+	}
+
+	return v, false
+}
+
+// GetRaw get empty interface
+func (p *PatchableValues) GetRaw(path string) interface{} {
```
nabokihms, line 44: Do we have a use case for this method? I think it will be rarely used.

nabokihms (approved):

pkg/module_manager/go_hook/patchable_values.go
yalosev: It is really rarely used. Just to avoid calls like x.Values.Values().(map[string]interface{}). But its cheaper to have it then to add again in future when we need it

PR #210, opened 2021-05-26
Improve PatchableValues Remove

yalosev:
#### Overview

Check that path exists on Values deletion

Check if path exists before deletion. If you call Remove
on absent path - it leads to error. Avoid this case

None


nabokihms (approved):
Make sense.

PR #224, opened 2021-06-16
Update shell-operator: k8s spec generator in the shell-operator

yalosev:
#### Overview

Replace instant object patcher with deferred one

To make the same behavior for shell-hooks and go-hooks we need to generate k8s object patches and execute them as batch after the hook has been finished

None

```release-note

```


nabokihms:
@yalosev, could you fix the conflicts? Otherwise, it is impossible to merge.

nabokihms (approved):

PR #234, opened 2021-09-06
Refactor helm3 communication

yalosev:
#### Overview

Helm release optimizations

The rollout of a new module is a bit slow cause of AbsentResources detection logic and helm3 binary calls.

 This patch:
- optimize absent resources figuring out
- add communication with helm3 as a library
- remove unused interface functions

NONE

```release-note

```


pkg/helm_resources_manager/resources_monitor.go
```diff
@@ -128,46 +128,145 @@ func (r *ResourcesMonitor) Resume() {
...
+
+// list all objects in ns and return names of all existent objects
+func (r *ResourcesMonitor) listResources(ctx context.Context, nsgvr namespacedGVR) (map[string]struct{}, error) {
+	var objList *unstructured.UnstructuredList
+	var err error
+	if nsgvr.Namespace == "_global_" {
+		objList, err = r.kubeClient.Dynamic().Resource(nsgvr.GVR).List(ctx, v1.ListOptions{})
+	} else {
+		objList, err = r.kubeClient.Dynamic().Resource(nsgvr.GVR).Namespace(nsgvr.Namespace).List(ctx, v1.ListOptions{})
+	}
```
diafour, line 259: See comment for buildGVRMap https://github.com/flant/addon-operator/pull/234/files#diff-e4cf48a8f46c87bfd44b6068348c77509ecaf426170f3b4ee1e07d2fc4ab0ff3R192

pkg/helm/helm3lib/helm3lib.go
```diff
@@ -0,0 +1,315 @@
...
+
+func (h *LibClient) UpgradeRelease(releaseName string, chartName string, valuesPaths []string, setValues []string, namespace string) error {
+	upg := action.NewUpgrade(actionConfig)
+	if namespace != "" {
+		upg.Namespace = namespace
+	}
+
+	upg.Install = true
+	upg.MaxHistory = int(options.HistoryMax)
+	upg.Timeout = options.Timeout
```
diafour, line 131: Just a thought... Shouldn't we enable Wait for helm3lib by default?

pkg/helm/helm3lib/helm3lib.go
```diff
@@ -0,0 +1,315 @@
...
+func (h *LibClient) GetReleaseValues(releaseName string) (utils.Values, error) {
+	gv := action.NewGetValues(actionConfig)
+	return gv.Run(releaseName)
+}
+
+func (h *LibClient) DeleteRelease(releaseName string) error {
+	h.LogEntry.Debugf("helm release '%s': execute helm uninstall", releaseName)
+
+	un := action.NewUninstall(actionConfig)
+	_, err := un.Run(releaseName)
```
diafour, line 202: helm3 package set Namespace for delete operation. I think ns is required here.

pkg/helm/helm3lib/helm3lib.go
```diff
@@ -0,0 +1,315 @@
...
+		return fmt.Errorf("helm upgrade failed: %s\n", err)
+	}
+	h.LogEntry.Infof("Helm upgrade for release '%s' with chart '%s' in namespace '%s' successful", releaseName, chartName, namespace)
+
+	return nil
+}
+
+func (h *LibClient) GetReleaseValues(releaseName string) (utils.Values, error) {
+	gv := action.NewGetValues(actionConfig)
+	return gv.Run(releaseName)
```
diafour, line 195: helm3 package set Namespace for 'get release values' operation. I think ns is required here.

diafour (changes requested):

pkg/helm/helm2/helm2.go
```diff
@@ -94,8 +94,8 @@ func (h *Helm2Client) Cmd(args ...string) (stdout string, stderr string, err err
 	return
 }
 
-// InitAndVersion runs helm init and helm version commands
-func (h *Helm2Client) InitAndVersion() error {
+// initAndVersion runs helm init and helm version commands
+func (h *Helm2Client) initAndVersion() error {
 	stdout, stderr, err := h.Cmd("init", "--client-only")
```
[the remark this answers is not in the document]
yalosev: Yes. Cmd function is used directly from /pkg/helm/helm_test.go. That test is skipped, so, maybe, it should be remove in the future.

pkg/helm_resources_manager/resources_monitor.go
```diff
@@ -128,46 +128,145 @@ func (r *ResourcesMonitor) Resume() {
...
 			Group:    apiRes.Group,
 			Version:  apiRes.Version,
 			Resource: apiRes.Name,
 		}
-		// Resources are filtered by metadata.name field. Object is considered absent if list is empty.
-		listOptions := v1.ListOptions{
-			FieldSelector: fields.OneTermEqualSelector("metadata.name", m.Name()).String(),
+		ns := m.Namespace(r.defaultNamespace)
+		if !apiRes.Namespaced {
+			ns = "_global_"
```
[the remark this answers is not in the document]
yalosev: I was confused with a previous version of the code. But it's really could be removed

pkg/helm/helm3lib/helm3lib.go
yalosev: Don't think so. Cause that way any broken deployment/daemonset/etc in CrashLoopBackoff status in any module could block the queue. And it will make deployment much more longer

pkg/helm/helm3lib/helm3lib.go
yalosev: Have no idea why but uninstall action does not have and Namespace attribute....

pkg/helm/helm3lib/helm3lib.go
yalosev: Same as for Uninstall... no ns here in attributes

pkg/helm/helm3lib/helm3lib.go
diafour: Oops. I thought wait is fail-fast.

pkg/helm_resources_manager/resources_monitor.go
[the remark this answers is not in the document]
diafour: That's ok then.

pkg/helm/helm2/helm2.go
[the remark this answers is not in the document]
diafour: That's ok then.

diafour (approved):

pkg/helm_resources_manager/resources_monitor.go
```diff
@@ -128,46 +127,139 @@ func (r *ResourcesMonitor) Resume() {
...
+				manifest:  mf,
+			}
+			return
 		}
 	}
+}
+
+// list all objects in ns and return names of all existent objects
+func (r *ResourcesMonitor) listResources(ctx context.Context, nsgvr namespacedGVR) (map[string]struct{}, error) {
+	objList, err := r.kubeClient.Dynamic().Resource(nsgvr.GVR).Namespace(nsgvr.Namespace).List(ctx, v1.ListOptions{})
```
nabokihms, line 252: if one list request fails, the whole upgrade checking fails too. Do we need to retry failed list requests in some cases?

pkg/helm_resources_manager/resources_monitor.go
```diff
@@ -128,46 +127,139 @@ func (r *ResourcesMonitor) Resume() {
...
+func (r *ResourcesMonitor) checkGVRManifests(ctx context.Context, wg *sync.WaitGroup, nsgvr namespacedGVR, manifests []manifest.Manifest, resC chan gvrManifestResult, concurrency chan struct{}) {
+	defer wg.Done()
+
+	concurrency <- struct{}{}
+	defer func() {
+		<-concurrency
+	}()
+
+	// if context is cancelled - return
+	if ctx.Err() != nil {
```
nabokihms, line 227: NIT: I think we need to check it before taking concurrency lock.

PR #239, opened 2021-09-21
[helm3] uninstall first revision in pending state

yalosev:
#### Overview

Fix helm3 pending state

If helm release has only 1 revision it could not be rollbacked. Uninstall it then

NONE

```release-note

```


pkg/helm/helm3lib/helm3lib.go
```diff
@@ -180,8 +180,14 @@ func (h *LibClient) UpgradeRelease(releaseName string, chartName string, valuesP
 		// looking through this issue you can found the common error: another operation (install/upgrade/rollback) is in progress
 		// and hints to fix it. In the future releases of helm they will handle sudden shutdown
 		if lr[0].Info.Status.IsPending() {
-			rb := action.NewRollback(actionConfig)
-			_ = rb.Run(lr[0].Name)
+			if lr[0].Version == 1 {
+				rb := action.NewUninstall(actionConfig)
```
nabokihms, line 184: I assume rb stays for RollBack, but there we use an uninstall action.

nabokihms (approved):

PR #259, opened 2021-11-08
Fix helm3lib upgrade

yalosev:
#### Overview

Fix release upgrade with helm3lib

Helm3Lib `Max` argument is not working. Anyway history client returns all releases and we have to fetch latest release manual from them.

NONE

```release-note

```


diafour (approved):
LGTM

PR #261, opened 2021-11-11
remove unused error

yalosev:
#### Overview

Remove unnecessary error when rolling back a release

This error is useless because we have a release upgrade immediately after a rollback. We don't care about rollback results. That behavior only leads to one more module restart during the error occurs

NONE
```release-note

```


diafour (approved):
ok!

PR #264, opened 2021-11-18
Fix resources_monitor goroutine leak

yalosev:
#### Overview

Remove goroutine leak

Non-buffered channel can stuck on chan send and we have non-finished leaking goroutines

NONE

```release-note

```


pkg/helm_resources_manager/resources_monitor.go
```diff
@@ -134,7 +134,7 @@ func (r *ResourcesMonitor) AbsentResources() ([]manifest.Manifest, error) {
 
 	concurrency := make(chan struct{}, 5)
 
-	resC := make(chan gvrManifestResult)
+	resC := make(chan gvrManifestResult, len(gvrMap))
```
nabokihms, line 137: Could you please add a comment on why do we need to set the length of the channel here to prevent the intention of changing this line in the future?

yalosev: done

nabokihms (approved):

PR #270, opened 2021-11-29
fix: enable locks for updating caches in ModuleManager.

diafour:
#### Overview

ModuleManager component stores caches with values and patches. This PR fixes the race when these caches update from parallel queues.

```release-note

```


nabokihms (commented):
I looked through it briefly, and everything seemed ok. Let's wait for the @yalosev's review before merging it.

diafour:
I've renamed a lock field and made it private.

yalosev (approved):

PR #300, opened 2022-02-28
Add flag to start scheduling 'schedule' hook before first converge done.

name212:
#### Overview

Revert starting schedule manager on start operator instead starting after first converge.

For go-hooks added flag `HookConfigSettings.EnableSchedulesOnStartup` for enable schedule hook without waiting finish converge.

By default all hooks (go and shell) will be running after converge. 

Some schedule hooks can run critical routines for operator (for ex, [deckhouse](https://github.com/deckhouse/deckhouse) uses them for auto-updating itself.)

Users can use schedule go-hooks before first converge.

```release-note

```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/davecgh/go-spew v1.1.1
 	github.com/evanphx/json-patch v4.11.0+incompatible
 	github.com/flant/kube-client v0.0.6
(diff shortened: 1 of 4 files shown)


diafour (approved):
Looks good. Now wait for https://github.com/flant/shell-operator/pull/360

PR #330, opened 2022-07-13
fix: ignore KubeConfig events before first converge

diafour:
#### Overview

Supporting changes for https://github.com/deckhouse/deckhouse/pull/1729

- Prevent queueing ConvergeModules while running global hooks at startup
- Add tracking last error for ModuleRun, ModuleDelete and ModuleHookRun
- Add GetModuleNames method

Global hook in 1729 changes ConfigMap and premature KubeConfigChanged event breaks enabled modules state.

```release-note

```

### DIFF
--- a/pkg/addon-operator/converge.go
+++ b/pkg/addon-operator/converge.go
@@ -10,11 +10,10 @@ import (
 )
 
 type ConvergeState struct {
-	Phase        ConvergePhase
-	FirstStarted bool
-	FirstDone    bool
-	StartedAt    int64
-	Activation   string
+	Phase         ConvergePhase
+	FirstRunPhase FirstConvergePhase
+	StartedAt     int64
+	Activation    string
 }
 
 type ConvergePhase string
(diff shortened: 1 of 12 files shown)


pkg/addon-operator/converge.go
```diff
@@ -27,9 +26,18 @@ const (
 	WaitAfterAll            ConvergePhase = "WaitAfterAll"
 )
 
+type FirstConvergePhase string
```
yalosev, line 29: We use it only for tracking an inner state. Don't you think we can use `int` here and `= iota` for initialization.
Comparing integers would be faster until we don't need string representation

pkg/module_manager/module_state.go
```diff
@@ -20,16 +25,42 @@ const (
...
+	s.hookErrorsLock.Lock()
+	defer s.hookErrorsLock.Unlock()
+
+	s.hookErrors[hookName] = err
+}
+
+// GetLastHookErr returns hook error.
+func (s *ModuleState) GetLastHookErr() error {
+	s.hookErrorsLock.Lock()
+	defer s.hookErrorsLock.Unlock()
```
yalosev, line 57: I guess we can use RLock here

yalosev (commented):
A couple of comments

pkg/addon-operator/converge.go
diafour: Done

pkg/module_manager/module_state.go
diafour: Done

yalosev (approved):

PR #335, opened 2022-09-07
forbid usage startup and kubernetes bindings together

yalosev:
#### Overview

Forbid usage OnStartup and Kubernetes bindings together

Hook with onStartup binding could not have kubernetes snapshots. Usually ability to declare them together leads to logic errors. Let's forbid it.

NONE

### DIFF
--- a/sdk/registry.go
+++ b/sdk/registry.go
@@ -55,6 +55,11 @@ func (h *HookRegistry) Add(hook go_hook.GoHook) {
 	h.m.Lock()
 	defer h.m.Unlock()
 
+	config := hook.Config()
+	if config.OnStartup != nil && len(config.Kubernetes) > 0 {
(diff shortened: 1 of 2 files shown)


sdk/registry.go
```diff
@@ -55,6 +55,11 @@ func (h *HookRegistry) Add(hook go_hook.GoHook) {
 	h.m.Lock()
 	defer h.m.Unlock()
 
+	config := hook.Config()
+	if config.OnStartup != nil && len(config.Kubernetes) > 0 {
+		panic("OnStartup and Kubernetes bindings cannot be used in a same hook")
```
diafour, line 60:
```suggestion
		panic("OnStartup hook always has binding context without Kubernetes snapshots. To prevent logic errors, don't use OnStartup and Kubernetes bindings in the same Go hook configuration.")
```

diafour (approved):

PR #367, opened 2023-03-02
Change rollback to secret delete

yalosev:
#### Overview

Delete secret with pending release before upgrade instead of rollback

NONE


yalosev:
nvm

PR #374, opened 2023-04-16
feat: Persist helm3lib Kubernetes clients

nabokihms:

This PR prevents instantiating the REST client multiple times. In our current configuration, helm discovery actively allocates memory, which makes GC executions more frequent. The full description can be found [here](https://github.com/kubernetes/cli-runtime/blob/4e17abf5754235e5f3adec6b5fcb71a1b6a504d7/pkg/genericclioptions/config_flags.go#L115-L118).

image
image
image

Benchmark results for getting the REST client with the previously used env getter and the new persistent getter:

```
goos: darwin
goarch: amd64
pkg: github.com/flant/addon-operator/pkg/helm/helm3lib
BenchmarkRESTMapper/Env_client-12                   1413            790653 ns/op          183872 B/op       1939 allocs/op
BenchmarkRESTMapper/Persistent_client-12        44701933             27.69 ns/op               0 B/op          0 allocs/op
```

Requires deep testing because this can lead to issues in cases when CRDs are deployed to the cluster, which means new resource types are dynamically added to a cluster.

Another way to solve the problem is to use a single client for all operations with Kubernetes.

```release-note
Persist helm3lib Kubernetes clients.
```

### DIFF
--- a/pkg/helm/helm.go
+++ b/pkg/helm/helm.go
@@ -33,7 +33,7 @@ func InitHelmClientFactory(kubeClient klient.Client) (*ClientFactory, error) {
 	case Helm3Lib:
 		log.Info("Helm3Lib detected. Use builtin Helm.")
 		factory.NewClientFn = helm3lib.NewClient
-		err = helm3lib.Init(&helm3lib.Options{
+		helm3lib.Init(&helm3lib.Options{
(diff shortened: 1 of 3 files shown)


yalosev (approved):
Sounds tasty :)

PR #375, opened 2023-04-18
fix: Speedup ApplyIgnoreNonExistentPaths jsonpatch method

nabokihms:
#### Overview
A CPU profile shows that the ApplyIgnoreNonExistentPaths method consumes a lot of resources to do json.Marshal -> Unmarshal. The problem is that addon-operator used the patch.Apply in a for loop to check the error of each patch. Unfortunately, every patch forced decoding and encoding of all the values.

With the approaches from this PR, the problem is avoided by using the library fork, where patch operations can be applied to the internal view of the JSON doc of the library - container.

image

Grafana panels after deploying addon-operator with the fix.
image

Benchmark:
```
goos: darwin
goarch: amd64
pkg: github.com/flant/addon-operator/pkg/utils
BenchmarkValuesPatchApply/Apply_Patch-12                    1286           1082385 ns/op          145288 B/op       1772 allocs/op
BenchmarkValuesPatchApply/Apply_Patch_Fast-12              10000            121489 ns/op           20177 B/op        454 allocs/op

```

```release-note
Speedup ApplyIgnoreNonExistentPaths jsonpatch method
```

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -36,6 +36,13 @@ linters-settings:
       - prefix(github.com/flant/)
   goimports:
     local-prefixes: github.com/flant/
+  depguard:
+    list-type: blacklist
+    include-go-root: true
+    packages:
+      - github.com/evanphx/json-patch
+    packages-with-error-message:
+      - github.com/evanphx/json-patch: "The 'github.com/evanphx/json-patch' package is superseded. Use pkg/utils/jsonpatch.go instead."
 issues:
   exclude:
   # Using underscores is a common practice, refactor in the future
--- a/pkg/utils/values_patch.go
+++ b/pkg/utils/values_patch.go
@@ -9,8 +9,6 @@ import (
 	"reflect"
 	"sort"
 	"strings"
-
-	jsonpatch "github.com/evanphx/json-patch"
 )
 
 type ValuesPatchType string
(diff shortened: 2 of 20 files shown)


pkg/utils/values_patch.go
```diff
@@ -24,19 +22,13 @@ type ValuesPatch struct {
 	Operations []*ValuesPatchOperation
 }
 
-func NewValuesPatch() *ValuesPatch {
```
yalosev, line 27: This method is used somewhere in the Deckhouse

.golangci.yaml
```diff
@@ -36,6 +36,13 @@ linters-settings:
       - prefix(github.com/flant/)
   goimports:
     local-prefixes: github.com/flant/
+  depguard:
+    list-type: blacklist
+    include-go-root: true
+    packages:
+      - github.com/evanphx/json-patch
```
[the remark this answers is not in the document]
nabokihms: Good catch! It is probably better to add the check that the mod file was cleaned with the `go mod tidy` command.

pkg/utils/values_patch.go
nabokihms: Reverted

yalosev (approved):

PR #377, opened 2023-04-20
Ignore module on absent symlink

yalosev:
#### Overview

Don't fall with a fatal error on an absent symlink

Multi-master installation could have broken symlinks. We should start anyway, just ignore broken module.

NONE

```release-note

```

### DIFF
--- a/pkg/module_manager/module_loader.go
+++ b/pkg/module_manager/module_loader.go
@@ -2,6 +2,7 @@ package module_manager
 
 import (
 	"fmt"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"regexp"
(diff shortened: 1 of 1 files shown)


pkg/module_manager/module_loader.go
```diff
@@ -92,8 +93,16 @@ func resolveDirEntry(dirPath string, entry os.DirEntry) (string, string, error)
 	// Check if entry is a symlink to a directory.
 	targetPath, err := resolveSymlinkToDir(dirPath, entry)
 	if err != nil {
+		if e, ok := err.(*fs.PathError); ok {
+			if e.Err.Error() == "no such file or directory" {
+				log.Warnf("Symlink target %q does not exist. Ignoring module", dirPath)
```
diafour, line 98: `%q` is handy, but it brings slashes in json logs: "..target \"modules/001-module\" does not exist...". 🥴 
 I prefer to use `'%s'` for logging.

nabokihms: This is a problem. JSON will be not valid if a string contains `"`. For example, `{"key": "test '%s' "}` will produce an invalid JSON document if %s is replaced with `"string in quotes"`.

nabokihms (approved):
Merge whenewer

PR #418, opened 2023-10-18
Update mergemap to reduce memory allocations

nabokihms:
#### Overview

The idea is to allocate less memory when the map is already a map[string]interface{} / Values type.

I decided to move the code to our repo, because
1. This is a single file with two functions, the size is not big enough 
2.  Code changes were required
3. To effectively check type aliases such as `Values`, the type should be imported and be used by the function (causing cyclo dependencies)

```
goos: darwin
goarch: amd64
pkg: github.com/peterbourgon/mergemap
                        │       old.txt       │                new.txt                │
                        │       sec/op        │    sec/op     vs base                 │
Mapify/Map_interface-12   45830274.000n ±  9%   2.558n ±  1%  -100.00% (p=0.000 n=20)
Mapify/Map_string-12             51.12m ± 10%   40.15m ± 16%   -21.46% (p=0.000 n=20)

                        │   old.txt    │                  new.txt                  │
                        │     B/op     │     B/op      vs base                     │
Mapify/Map_interface-12   15.49Mi ± 0%    0.00Mi ± 0%  -100.00% (p=0.000 n=20)
Mapify/Map_string-12      15.48Mi ± 0%   10.31Mi ± 0%   -33.39% (p=0.000 n=20)

                        │   old.txt   │                 new.txt                  │
                        │  allocs/op  │  allocs/op   vs base                     │
Mapify/Map_interface-12   203.9k ± 0%     0.0k ± 0%  -100.00% (p=0.000 n=20)
Mapify/Map_string-12      203.9k ± 0%   201.7k ± 0%    -1.11% (p=0.000 n=20)
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,8 +14,8 @@ require (
 	github.com/go-openapi/validate v0.19.12
 	github.com/hashicorp/go-multierror v1.1.1
 	github.com/kennygrant/sanitize v1.2.4
+	github.com/nabokihms/mergemap v0.0.1
 	github.com/onsi/gomega v1.27.10
-	github.com/peterbourgon/mergemap v0.0.1
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.16.0
 	github.com/segmentio/go-camelcase v0.0.0-20160726192923-7085f1e3c734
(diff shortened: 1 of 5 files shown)


nabokihms:
The only concern is that without copying the map the function can mutate some values. For now, I do not see any problems in a living cluster.

yalosev (approved):

PR #432, opened 2023-12-06
[addon-operator] add module reregister

miklezzzz:
#### Overview
Changes:
A lot of changes related to starting/restarting/disabling a module without restarting addon-operator.

With this PR we can provide a way of reloading (restarting its life cycle like disabling and deregistering its hooks, reloading its configuration and re-run the module)  a single module without restarting addon-operator.

### DIFF
--- a/pkg/kube_config_manager/kube_config_manager.go
+++ b/pkg/kube_config_manager/kube_config_manager.go
@@ -78,7 +78,7 @@ func (kcm *KubeConfigManager) IsModuleEnabled(moduleName string) bool {
 		return false
 	}
 
-	return *moduleConfig.IsEnabled
+	return moduleConfig.IsEnabled != nil && *moduleConfig.IsEnabled
 }
 
 func (kcm *KubeConfigManager) Init() error {
@@ -119,12 +119,12 @@ func (kcm *KubeConfigManager) KubeConfigEventCh() chan config.KubeConfigEvent {
 
 // UpdateModuleConfig updates a single module config
 func (kcm *KubeConfigManager) UpdateModuleConfig(moduleName string) error {
-	newConfig, err := kcm.backend.LoadConfig(kcm.ctx)
+	newModuleConfig, err := kcm.backend.LoadConfig(kcm.ctx, moduleName)
 	if err != nil {
 		return err
 	}
 
-	if moduleConfig, found := newConfig.Modules[moduleName]; found {
+	if moduleConfig, found := newModuleConfig.Modules[moduleName]; found {
 		if kcm.knownChecksums != nil {
 			kcm.knownChecksums.Set(moduleName, moduleConfig.Checksum)
 		}
--- a/pkg/kube_config_manager/kube_config_manager.go
+++ b/pkg/kube_config_manager/kube_config_manager.go
@@ -117,6 +117,24 @@ func (kcm *KubeConfigManager) KubeConfigEventCh() chan config.KubeConfigEvent {
 	return kcm.configEventCh
 }
 
+// UpdateModuleConfig updates a single module config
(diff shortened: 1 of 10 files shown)


pkg/module_manager/module_manager.go
```diff
@@ -1141,3 +1225,41 @@ func loadStaticValues(moduleName, modulePath string) (utils.Values, error) {
...
+}
+
+// modulesWithPendingModuleRun returns names of all modules in pending
+// ModuleRun tasks. First task in queue considered not pending and is ignored.
+func modulesWithPendingModuleRun(q *queue.TaskQueue) map[string]struct{ doStartup bool } {
+	if q == nil {
+		return nil
+	}
+
+	modules := make(map[string]struct{ doStartup bool })
```
yalosev, line 1247: This map looks so complicated. Why we couldn't use `map[string]bool` ?

miklezzzz: I also find this piece of code debatable, for several realons:
* it's almost a complete copy of https://github.com/flant/addon-operator/blob/a9a25d81587258f0196cf315cf0ecd064055ec73/pkg/addon-operator/queue.go#L11-L45, I decided to copy it just in favor of not cluttering the PR with a lot of refactoring (it may seem it makes sense to move some queue-related functions to a dedicated util file, may it not?)
* as of map[string]struct - queues tasks have a lot different meta, at this point we need only DoStartup flag, while next time we could need something else, this is why I decided to have a go at using an anonymous struct

pkg/module_manager/loader/loader.go
```diff
@@ -6,4 +6,5 @@ import (
 
 type ModuleLoader interface {
 	LoadModules() ([]*modules.BasicModule, error)
+	ReloadModule(string, string) (*modules.BasicModule, error)
```
yalosev, line 9:
```suggestion
	ReloadModule(moduleName string, modulePath string) (*modules.BasicModule, error)
```

pkg/kube_config_manager/kube_config_manager.go
```diff
@@ -72,6 +72,15 @@ func NewKubeConfigManager(ctx context.Context, bk backend.ConfigHandler, runtime
 	}
 }
 
+func (kcm *KubeConfigManager) IsModuleEnabled(moduleName string) bool {
+	moduleConfig, found := kcm.currentConfig.Modules[moduleName]
+	if !found {
+		return false
+	}
+
+	return *moduleConfig.IsEnabled
```
yalosev, line 81: `IsEnabled` could be nil. We have to add a check

pkg/kube_config_manager/kube_config_manager.go
```diff
@@ -108,6 +117,24 @@ func (kcm *KubeConfigManager) KubeConfigEventCh() chan config.KubeConfigEvent {
 	return kcm.configEventCh
 }
 
+// UpdateModuleConfig updates a single module config
+func (kcm *KubeConfigManager) UpdateModuleConfig(moduleName string) error {
+	newConfig, err := kcm.backend.LoadConfig(kcm.ctx)
```
yalosev, line 122: I think we have to add variadic parameter `moduleName ...string` to the LoadConfig function, to have ability to reload only desired ModuleConfig

pkg/kube_config_manager/kube_config_manager.go
miklezzzz: done

pkg/kube_config_manager/kube_config_manager.go
miklezzzz: done

yalosev (approved):

PR #444, opened 2023-12-26
validate all enabled modules

miklezzzz:
#### Overview

This pr should fix the situation when dynamically enabled modules don't get their config values applied if `enabled: true` isn't set in their configuration. Also, there were several other problems to solve like: module configs for dynamically enabled modules aren't applied at startup.
Relates to https://github.com/deckhouse/deckhouse/pull/7036.

There is a problem with dynamically enabled modules (that are launched by global hooks) not getting their kube config values applied  if `enabled: true` isn't set explicitly. It happens because HandleNewKubeConfig method doesn't count dynamically enabled modules as modules enabled by config which is partially true (enabled isn't set), but these modules still should have their config values applied and recalculated each time their configs change if `enabled: false` isn't set.

### DIFF
--- a/pkg/module_manager/models/modules/values_storage.go
+++ b/pkg/module_manager/models/modules/values_storage.go
@@ -216,7 +216,8 @@ func (vs *ValuesStorage) SaveConfigValues(configV utils.Values) {
 	}
 }
 
-// GenerateNewConfigValues generated new config values, based on static and openapi specs
+// GenerateNewConfigValues generated new config values, based on static and config values. Additionally, if validate is true, it validates
+// two versions of resulting values: original (without openapi defaults) and with openapi defaults for consistency.
 // this method always makes a copy of the result values to prevent pointers shallow copying
 func (vs *ValuesStorage) GenerateNewConfigValues(configV utils.Values, validate bool) (utils.Values, error) {
 	vs.lock.Lock()
@@ -227,9 +228,6 @@ func (vs *ValuesStorage) GenerateNewConfigValues(configV utils.Values, validate 
 		// Init static values
 		vs.staticConfigValues,
 
-		// defaults from openapi
-		vs.openapiDefaultsTransformer(validation.ConfigValuesSchema),
-
 		// User configured values (ConfigValues)
 		configV,
 	)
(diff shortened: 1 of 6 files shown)


pkg/module_manager/models/modules/values_storage.go
```diff
@@ -242,6 +240,19 @@ func (vs *ValuesStorage) GenerateNewConfigValues(configV utils.Values, validate
...
 			return nil, err
 		}
+
+		mergedWithOpenapiDefault := mergeLayers(
+			utils.Values{},
+			vs.staticConfigValues,
+			vs.openapiDefaultsTransformer(validation.ConfigValuesSchema),
+			configV,
+		)
+		mergedWithOpenapiDefault = mergedWithOpenapiDefault.Copy()
```
yalosev, line 250: I don't think that we have to copy this, because we don't return this variable, only use it for validation

pkg/module_manager/module_manager.go
```diff
@@ -217,8 +217,12 @@ func (mm *ModuleManager) HandleNewKubeConfig(kubeConfig *config.KubeConfig) (*Mo
 
 	// Get map of enabled modules after KubeConfig changes.
 	newEnabledByConfig := mm.calculateEnabledModulesByConfig(kubeConfig)
+	allModules := make(map[string]struct{})
```
yalosev, line 220:
```suggestion
	allModules := make(map[string]struct{}, len(mm.modules.NamesInOrder()))
```

pkg/module_manager/models/modules/values_storage.go
yalosev: also I don't think that we need to validate `merged` values without openapi defaults. It seems like double job. Please remove this
```
err := vs.validateConfigValues(merged)
```

pkg/module_manager/module_manager.go
miklezzzz: done

pkg/module_manager/models/modules/values_storage.go
miklezzzz: done

yalosev (approved):

PR #446, opened 2024-01-10
Added logs before os.Exit(1) in main.go

akaitux:
#### Overview
Added logs before os.Exit(1) with error in main.go

If you use the operator as is and do not add the /global-hooks directory to the image, the code exits with exit code 1 without any error messages.

### DIFF
--- a/cmd/addon-operator/main.go
+++ b/cmd/addon-operator/main.go
@@ -54,11 +54,13 @@ func main() {
 
 			err := operator.Setup()
 			if err != nil {
+                fmt.Printf("Setup is failed: %s\n", err)
 				os.Exit(1)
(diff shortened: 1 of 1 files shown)


nabokihms (approved):
Thank you for your contribution!

diafour:
Oops you've broken JSON logging =) `fmt.Printf` should be replaced with `log.Error`.

akaitux:
@diafour Sorry, I specifically made an issue so that someone from the team would add these changes, since I don’t know the entire infrastructure. Should I make another pr or will you fix it yourself?

nabokihms:
This is not a problem, because this is the first error.

PR #460, opened 2024-03-18
[module-manager] Deployed and installed external modules aren't deleted if relevant modules configs are deleted

libmonsoon-dev:
#### Overview
Added removal of a module when deleting its configuration

https://github.com/deckhouse/deckhouse/issues/7734


pkg/kube_config_manager/kube_config_manager.go
```diff
@@ -243,8 +247,10 @@ func (kcm *KubeConfigManager) handleConfigEvent(obj config.Event) {
 			kcm.logEntry.Infof("Module section deleted: %+v", moduleName)
 			moduleCfg.DropValues()
 			moduleCfg.Checksum = moduleCfg.ModuleConfig.Checksum()
+			moduleCfg.IsEnabled = pointer.Bool(false)
```
yalosev, line 250: It's not valid for embedded modules

yalosev:
Will realize it in the other CRD

PR #464, opened 2024-03-21
fix camelCase to kebab-case conversion

miklezzzz:
#### Overview
The camelCase to kebab-case string conversion function is updated to use "github.com/ettle/strcase" package instead of local implementation. It should allow converting module names containing digits more correctly.

It updates the way we convert camelCased strings to kebab-cased. At now, if we have something like 'l2-load-balancer' in kebab-case, after converting to camelCase and back it ends up with 'l-2-load-balancer'.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.19
 require (
 	github.com/davecgh/go-spew v1.1.1
 	github.com/distribution/distribution/v3 v3.0.0-20221208165359-362910506bc2
+	github.com/ettle/strcase v0.2.0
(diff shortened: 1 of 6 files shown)


miklezzzz:
here is a diff output between values `before` and `after` the changes
image

miklezzzz:
I think I can't comprehend all possible consequences of these changes, need you to spend a couple of minutes thinking through a possible worst case scenario :) @nabokihms @yalosev

yalosev (approved):

PR #465, opened 2024-04-02
Scheduler

yalosev:
#### Overview

- Reworking the mechanism of detecting which modules are enabled/disabled and by which components;
- RunModuleWithNewStaticValues method was rewritten to RunModuleWithNewOpenAPISchema (and it's logic, from updating a module's static values to updating its schema storage;
- A `module status changed` event is sent if the module's enabled status wasn't changed, but the reason why module is enabled/disabled has changed.

This pr introduces a new sub-component of Module Manager - Module Scheduler, which aim is to decide which Module Manager's modules should be enabled or disabled by means of calculating a graph of the modules (DAG), traversing the graph and applying various module Filters (called Extenders). Currently, there are four basic extenders to filter modules' status: 
- static (based on static values files)
- dynamic (based on catching global hooks' patches)
- kube-config (based on the values provided in the comfigMap)
- script-enabled (based on the results of running modules' enabled scripts)
The order extenders applied is up to the user, yet a default list of applied extenders is provided. It's up to the user to add any new extenders, following provided extender interface.

Also, there are numerous changes to other components of the operator in the pr to suit the new graph-based model.

### DIFF
--- a/pkg/module_manager/scheduler/scheduler.go
+++ b/pkg/module_manager/scheduler/scheduler.go
@@ -1,16 +1,18 @@
 package scheduler
 
 import (
+	"bufio"
+	"bytes"
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
-	"os"
+	"image"
 	"strings"
 	"sync"
 
 	"github.com/dominikbraun/graph"
 	"github.com/dominikbraun/graph/draw"
+	"github.com/goccy/go-graphviz"
 	log "github.com/sirupsen/logrus"
 
 	"github.com/flant/addon-operator/pkg/module_manager/scheduler/extenders"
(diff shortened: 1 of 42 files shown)


pkg/module_manager/scheduler/scheduler.go
```diff
@@ -0,0 +1,459 @@
...
+	root                *node.Node
+	enabledModules      *[]string
+	retrospectiveStatus map[extenders.ExtenderName]map[string]bool
+}
+
+type Extender interface {
+	Name() extenders.ExtenderName
+	Filter(module node.ModuleInterface) (*bool, error)
+	IsShutter() bool
+	IsResettable() bool
```
yalosev, line 32: Do we really need that?
We can just create an empty Reset function which do nothing
```
func (a SomeExtender) Reset() error {
  return nil
}
```

miklezzzz: yeah, my bad, fixed it

miklezzzz (approved):

PR #469, opened 2024-04-15
Provide a way of updating a module's static values

miklezzzz:
#### Overview
- Fixed LoadModule() logic (previously, it didn't load static values from the module's parent directory);
- RunModuleWithNewStaticValues method is introduced to renew the module's static values and recalculate resulting values of the module;
- Additional methods for BasicModule are added: to get the module's static values and to set(and recalculate) resulting values;
- PushRunModuleTask method is updated and it is possible to define if start-up hooks should be run on the module's run.

In two words, this PR provides additional API to work with modules' static values so as not only do a module's static values get set on the operator's startup, but they also can be updated and recalculated later. 
Relates to https://github.com/deckhouse/deckhouse/pull/8067

### DIFF
--- a/pkg/module_manager/models/modules/basic.go
+++ b/pkg/module_manager/models/modules/basic.go
@@ -120,6 +120,17 @@ func (bm *BasicModule) ResetState() {
 	}
 }
 
+// GetStaticValues returns the module's static values
+func (bm *BasicModule) GetStaticValues() utils.Values {
+	return bm.valuesStorage.staticConfigValues
+}
+
+// SetStaticValuesAndRecalculate sets the module's static values and recalculate the resulting values
+func (bm *BasicModule) SetStaticValuesAndRecalculate(values utils.Values) error {
+	bm.valuesStorage.staticConfigValues = values
+	return bm.valuesStorage.CommitValues()
+}
+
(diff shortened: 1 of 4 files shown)


pkg/module_manager/models/modules/basic.go
```diff
@@ -120,6 +120,17 @@ func (bm *BasicModule) ResetState() {
 	}
 }
 
+// GetStaticValues returns the module's static values
+func (bm *BasicModule) GetStaticValues() utils.Values {
+	return bm.valuesStorage.staticConfigValues
```
yalosev, line 125: shouldn't we get a lock here?

pkg/module_manager/models/modules/basic.go
```diff
@@ -120,6 +120,17 @@ func (bm *BasicModule) ResetState() {
...
+}
+
+// SetStaticValuesAndRecalculate sets the module's static values and recalculate the resulting values
+func (bm *BasicModule) SetStaticValuesAndRecalculate(values utils.Values) error {
+	bm.valuesStorage.staticConfigValues = values
```
yalosev, line 130: also this operation without a lock could be dangerous

PR #470, opened 2024-04-17
[addon-operator] Get rid of ValuesValidator structure

libmonsoon-dev:
#### Overview
Validation logic has been moved to the module object

https://github.com/deckhouse/deckhouse/issues/6884


pkg/module_manager/module_manager_hooks.go
```diff
@@ -15,15 +15,22 @@ import (
 	sh_op_types "github.com/flant/shell-operator/pkg/hook/types"
 )
 
-func (mm *ModuleManager) loadGlobalValues() ( /* global values */ utils.Values /* enabled modules */, map[string]struct{}, error) {
+type globalValues struct {
+	globalValues   utils.Values
+	enabledModules map[string]struct{}
+	configBytes    []byte // config-values.yaml
```
yalosev, line 21: NIT: I guess we can rename it to `configSchema` or something like that. Because name `...Bytes` is not informative

pkg/module_manager/module_manager_hooks.go
```diff
@@ -80,8 +84,13 @@ func (mm *ModuleManager) registerGlobalModule(globalValues utils.Values) error {
...
 	}
 
-	gm := modules.NewGlobalModule(mm.GlobalHooksDir, globalValues, mm.ValuesValidator, &dep)
+	gm, err := modules.NewGlobalModule(mm.GlobalHooksDir, globalValues, &dep, configBytes, valuesBytes)
+	if err != nil {
+		return fmt.Errorf("new global module: %w", err)
+	}
+
 	mm.global = gm
+	log.Infof(gm.GetSchemaStorage().GlobalSchemasDescription())
```
yalosev, line 93: Is it some debug info?

pkg/module_manager/models/modules/basic.go
```diff
@@ -862,6 +868,68 @@ func (bm *BasicModule) GetModuleError() error {
 	return bm.state.lastModuleErr
 }
 
+func (bm *BasicModule) GetValuesStorage() *ValuesStorage {
+	return bm.valuesStorage
+}
+
+func (bm *BasicModule) GetSchemaStorage() *validation.SchemaStorage {
```
yalosev, line 875: This method is not used anywhere. Do we need it?

pkg/module_manager/models/modules/global.go
```diff
@@ -499,3 +507,7 @@ func (gm *GlobalModule) searchGlobalGoHooks() ([]*kind.GoHook, error) {
 
 	return goHooks, nil
 }
+
+func (gm *GlobalModule) GetSchemaStorage() *validation.SchemaStorage {
```
yalosev, line 511: Also here - don't need it

yalosev (approved):

pkg/module_manager/module_manager_hooks.go
libmonsoon-dev: Yes, but this line was [there](https://github.com/flant/addon-operator/pull/470/files/dae21fbd4c189252ef4a591f88b9cfa39c2b921f#diff-e660fb6fddb6a937f54c49904c858dfd52b579e5a4a9fa414e046976b75a6fb4L73) earlier, so I decided to leave it

pkg/module_manager/module_manager_hooks.go
libmonsoon-dev: Fixed

PR #484, opened 2024-06-27
[addon-operator] feat: change behavior for statusValues

juev:
#### Overview

When building configuration files, do not take values ​​from values.yaml files

The modules are configured in one place, uniformly, through value management. There is no need to control the process of merging changes at different levels.

Closed: [deckhouse#8810](https://github.com/deckhouse/deckhouse/issues/8810)

BREAKING CHANGES:
`values.yaml` in the root of a module are no longer considered as ConfigValues for a module, only as Values

### DIFF
--- a/pkg/module_manager/models/modules/basic.go
+++ b/pkg/module_manager/models/modules/basic.go
@@ -77,7 +77,7 @@ func NewBasicModule(name, path string, order uint32, staticValues utils.Values, 
 	}, nil
 }
 
-// WithDependencies unject module dependencies
+// WithDependencies inject module dependencies
 func (bm *BasicModule) WithDependencies(dep *hooks.HookExecutionDependencyContainer) {
 	bm.dc = dep
 }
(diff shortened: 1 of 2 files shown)


pkg/module_manager/models/modules/basic.go
```diff
@@ -627,7 +627,7 @@ func (bm *BasicModule) executeHook(h *hooks.ModuleHook, bindingType sh_op_types.
 		bm.dc.MetricStorage.HistogramObserve("{PREFIX}module_hook_run_user_cpu_seconds", hookResult.Usage.User.Seconds(), metricLabels, nil)
 		bm.dc.MetricStorage.GaugeSet("{PREFIX}module_hook_run_max_rss_bytes", float64(hookResult.Usage.MaxRss)*1024, metricLabels)
 	}
-	if err != nil {
+	if err != nil || hookResult == nil {
```
yalosev, line 630: One line below we have the check
```
if hookResult != nil && ...
```
It's a bit weird, to pass the check `hookResult == nil` and next line check `hookResult != nil`

yalosev (approved):

PR #497, opened 2024-08-19
feat: applyCRDs on start main queue

juev:
#### Overview

We want to get rid from crd modules, like operator-prometheus-crd. Instead of that we have to discover CRDs and install them automatically on startup.

Fixed: [#9387](https://github.com/deckhouse/deckhouse/issues/9387)

When an error occurs during the crds application stage, we record them, but repeat them endlessly.

### DIFF
--- a/docs/src/RUNNING.md
+++ b/docs/src/RUNNING.md
@@ -67,9 +67,11 @@ dev_cluster_live_ticks 32
 ...
 ```
 
-**ADDON_OPERATOR_CRD_EXTRA_LABELS** – string with CRDs label selectors. Default is `heritage=addon-operator`.
+**ADDON_OPERATOR_CRD_EXTRA_LABELS** – string with CRDs label pairs.  
+For example: `heritage=my-app,scope=extra`  
+Default is `heritage=addon-operator`.
 
-**ADDON_OPERATOR_CRD_FILTERS** – String of filters for the CRD, separated by commas. Default is `doc-,_`.
+**ADDON_OPERATOR_CRD_FILTER_PREFIXES** – String of filters for the CRD, separated by commas. Default is `doc-,_`.
 
 ### Kubernetes client settings
 
--- a/docs/src/RUNNING.md
+++ b/docs/src/RUNNING.md
@@ -67,6 +67,10 @@ dev_cluster_live_ticks 32
 ...
 ```
 
+**ADDON_OPERATOR_CRD_EXTRA_LABELS** – string with CRDs label selectors. Default is `heritage=addon-operator`.
+
+**ADDON_OPERATOR_CRD_FILTERS** – String of filters for the CRD, separated by commas. Default is `doc-,_`.
+
 ### Kubernetes client settings
 
 **KUBE_CONFIG** — a path to a kubernetes client config (~/.kube/config)
--- a/pkg/addon-operator/bootstrap.go
+++ b/pkg/addon-operator/bootstrap.go
@@ -14,7 +14,7 @@ import (
 
 // Bootstrap inits all dependencies for a full-fledged AddonOperator instance.
 func (op *AddonOperator) bootstrap() error {
-	log.Infof(sh_app.AppStartMessage)
+	log.Info(sh_app.AppStartMessage)
 
 	log.Infof("Search modules in: %s", app.ModulesDir)
 
--- a/pkg/addon-operator/ensure_crds.go
+++ b/pkg/addon-operator/ensure_crds.go
@@ -6,9 +6,7 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"path/filepath"
 	"reflect"
-	"strings"
 
 	"github.com/hashicorp/go-multierror"
 	v1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
@@ -66,10 +64,6 @@ func (cp *CRDsInstaller) Run(ctx context.Context) *multierror.Error {
 	result := new(multierror.Error)
 
 	for _, crdFilePath := range cp.crdFilesPaths {
-		if match := strings.HasPrefix(filepath.Base(crdFilePath), "doc-"); match {
-			continue
-		}
-
 		err := cp.processCRD(ctx, crdFilePath)
 		if err != nil {
 			err = fmt.Errorf("error occurred during processing %q file: %w", crdFilePath, err)
--- a/pkg/addon-operator/ensure_crds.go
+++ b/pkg/addon-operator/ensure_crds.go
@@ -1,0 +1,225 @@
+package addon_operator
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"reflect"
+	"strings"
+
+	"github.com/hashicorp/go-multierror"
+	v1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	apimachineryv1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	apimachineryYaml "k8s.io/apimachinery/pkg/util/yaml"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/util/retry"
+
+	"github.com/flant/addon-operator/pkg/module_manager/models/modules"
+	"github.com/flant/addon-operator/sdk"
+	"github.com/flant/kube-client/client"
+)
+
+var crdGVR = schema.GroupVersionResource{
+	Group:    "apiextensions.k8s.io",
+	Version:  "v1",
+	Resource: "customresourcedefinitions",
+}
+
+func (op *AddonOperator) EnsureCRDs(module *modules.BasicModule) error {
+	crdsGlob := filepath.Join(module.GetPath(), "crds", "*.yaml")
+	result := new(multierror.Error)
+	cp, err := NewCRDsInstaller(op.KubeClient(), crdsGlob)
+	if err != nil {
+		result = multierror.Append(result, err)
+		return result
+	}
+
+	if cp == nil {
+		return nil
+	}
+
+	return cp.Run(context.TODO()).ErrorOrNil()
+}
+
+// CRDsInstaller simultaneously installs CRDs from specified directory
+type CRDsInstaller struct {
+	k8sClient    dynamic.Interface
+	crdFilesPath []string
+	buffer       []byte
+
+	// concurrent tasks to create resource in a k8s cluster
+	k8sTasks *multierror.Group
+}
+
+func (cp *CRDsInstaller) Run(ctx context.Context) *multierror.Error {
+	result := new(multierror.Error)
+
+	for _, crdFilePath := range cp.crdFilesPath {
+		if match := strings.HasPrefix(filepath.Base(crdFilePath), "doc-"); match {
+			continue
+		}
+
+		err := cp.processCRD(ctx, crdFilePath)
+		if err != nil {
+			err = fmt.Errorf("error occurred during processing %q file: %w", crdFilePath, err)
+			result = multierror.Append(result, err)
+			continue
+		}
+	}
+
+	errs := cp.k8sTasks.Wait()
+	if errs.ErrorOrNil() != nil {
+		result = multierror.Append(result, errs.Errors...)
+	}
+
+	return result
+}
+
+func (cp *CRDsInstaller) processCRD(ctx context.Context, crdFilePath string) error {
+	crdFileReader, err := os.Open(crdFilePath)
+	if err != nil {
+		return err
+	}
+	defer crdFileReader.Close()
+
+	crdReader := apimachineryYaml.NewDocumentDecoder(crdFileReader)
+
+	for {
+		n, err := crdReader.Read(cp.buffer)
+		if err != nil {
+			if err == io.EOF {
+				break
+			}
+
+			return err
+		}
+
+		data := cp.buffer[:n]
+		if len(data) == 0 {
+			// some empty yaml document, or empty string before separator
+			continue
+		}
+		rd := bytes.NewReader(data)
+		err = cp.putCRDToCluster(ctx, rd, n)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (cp *CRDsInstaller) putCRDToCluster(ctx context.Context, crdReader io.Reader, bufferSize int) error {
+	var crd *v1.CustomResourceDefinition
+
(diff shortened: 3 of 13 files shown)


pkg/addon-operator/operator.go
```diff
@@ -1481,6 +1481,17 @@ func (op *AddonOperator) HandleModuleRun(t sh_task.Task, labels map[string]strin
 				logEntry.Debugf("ModuleRun '%s' phase", baseModule.GetPhase())
 
 				treg := trace.StartRegion(context.Background(), "ModuleRun-OnStartup")
+				defer treg.End()
+
+				// Ensure modules CRDs
+				if err := op.EnsureCRDs(baseModule); err != nil {
+					op.ModuleManager.UpdateModuleLastErrorAndNotify(baseModule, err)
+					res.Status = queue.Fail
```
juev, line 1489: we can have different types of errors, conflict, or apiserver unavailability, or validation errors. Since it is difficult to determine what type of error it is, we mark the task as erroneous and repeat it until the problem is solved. Yes, this leads to the fact that in case of problems with crd validation we will repeat the operation infinitely

pkg/addon-operator/bootstrap.go
```diff
@@ -14,7 +14,7 @@ import (
 
 // Bootstrap inits all dependencies for a full-fledged AddonOperator instance.
 func (op *AddonOperator) bootstrap() error {
-	log.Infof(sh_app.AppStartMessage)
+	log.Info(sh_app.AppStartMessage)
```
juev, line 17: All edits to Infof -> Info or Errorf -> Error are made to fix golangci-lint errors that occur when running locally.

juev:
# How I tested

## Correct module

I created a [separate branch](https://github.com/deckhouse/deckhouse/tree/feature/9387-install-crd-on-startup) in the DH project, where I used this version of addon-operator. In this branch, I removed the call to the ensure_crds hook from a number of modules. All tests were performed with the working multitenancy-manager module.

Then I deleted the module-related crds from the dev cluster. Restarted the cluster.

After the restart was completed, I saw that the module crds were restored.

Then I disabled the module. Removing its crd from the dev cluster. And restarting the controller.

After the restart, the module remained disabled, and its crd was also missing, i.e. not processed.

## Failed module

I then made changes to the DH branch that resulted in an incorrect crd file for the module I was using.

After removing crd from the dev cluster and restarting the controller, we get a stop of the task processing queue indicating the error that occurred.

Screenshot 2024-08-26 at 15 39 02
Screenshot 2024-08-26 at 16 48 44
Screenshot 2024-08-26 at 17 20 51

The task queue will be stopped until the error in the crd file is corrected.
We can also manually disable the module. This will restart the task queue.

miklezzzz (commented):
I could be wrong, but if my memory serves me well, the idea was to install all module's CRDs in one step before starting the modules converge process. In this PR, we apply CRDs gradually along with processing HandleModuleRun tasks. It still can lead to a problem when a module with weight (order) 900 has some custom CRDS to apply to the cluster and we have some validating webhook for the module's CRDs to catch. Thus, the webhook tries to prepare a handler, but the CRD's aren't installed in the cluster yet, causing the handler to fail.

juev:
> I could be wrong, but if my memory serves me well, the idea was to install all module's CRDs in one step before starting the modules converge process. In this PR, we apply CRDs gradually along with processing HandleModuleRun tasks. It still can lead to a problem when a module with weight (order) 900 has some custom CRDS to apply to the cluster and we have some validating webhook for the module's CRDs to catch. Thus, the webhook tries to prepare a handler, but the CRD's aren't installed in the cluster yet, causing the handler to fail.

Good remark. I missed it. I’ll correct it.

juev:
# How I tested

## Ensure CRDs

First, I checked that EnsureCRDs starts before all modules start. To do this, I looked at the DH logs. Example of execution lines. The lines were before the first module started.

Screenshot 2024-08-28 at 18 34 46

Then I enabled the module that had the ensure_crds hook removed. After the queue had passed, all the necessary crds were in the cluster.

## Failed CRDs

Then I checked how the queue works in case of broken CRDs in the modules.

To do this, I created broken CRDs in the DH branch and enabled this module.

Screenshot 2024-08-28 at 18 38 46

The queue in this case arises at the stage of using broken CRDs. We do not reach the start of the modules.

The same thing happens when you restart the controller.

If after this we turn off the module, the queue passes, but we see an active task for applying CRDs with broken files.

Which results in the queue moving very slowly.

Screenshot 2024-08-28 at 18 44 38

Restarting the controller removes this task and the queue goes through without problems quite quickly.

juev:
Minor update to log output (error example):

Screenshot 2024-08-29 at 12 05 41

pkg/addon-operator/operator.go
```diff
@@ -1481,6 +1507,7 @@ func (op *AddonOperator) HandleModuleRun(t sh_task.Task, labels map[string]strin
 				logEntry.Debugf("ModuleRun '%s' phase", baseModule.GetPhase())
 
 				treg := trace.StartRegion(context.Background(), "ModuleRun-OnStartup")
+				defer treg.End()
```
miklezzzz, line 1510: not sure about it, this way it would trace everything from OnStartup stage till the end

pkg/addon-operator/ensure_crds.go
```diff
@@ -0,0 +1,225 @@
...
+	err = sdk.FromUnstructured(o, &crd)
+	if err != nil {
+		return nil, err
+	}
+
+	return crd, nil
+}
+
+// NewCRDsInstaller creates new installer for CRDs
+// crdsGlob example: "/deckhouse/modules/002-deckhouse/crds/*.yaml"
```
miklezzzz, line 205:
```suggestion
// crdsGlob example: "/modules/nginx/crds/*.yaml"
```

pkg/addon-operator/ensure_crds.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	if crd.APIVersion != v1.SchemeGroupVersion.String() && crd.Kind != "CustomResourceDefinition" {
+		return fmt.Errorf("invalid CRD document apiversion/kind: '%s/%s'", crd.APIVersion, crd.Kind)
+	}
+
+	if len(crd.ObjectMeta.Labels) == 0 {
+		crd.ObjectMeta.Labels = make(map[string]string, 1)
+	}
+	crd.ObjectMeta.Labels["heritage"] = "deckhouse"
```
yalosev, line 137: We should set it via env variables

Something like
`os.Getenv("ADDON_OPERATOR_CRD_EXTRALABELS")`
and set it in the deckhouse `ADDON_OPERATOR_CRD_EXTRALABELS=heritage=deckhouse,foo=bar`

pkg/addon-operator/ensure_crds.go
```diff
@@ -0,0 +1,225 @@
...
+			}
+
+			return err
+		}
+
+		if existCRD.Spec.Conversion != nil {
+			crd.Spec.Conversion = existCRD.Spec.Conversion
+		}
+
+		if existCRD.GetObjectMeta().GetLabels()["heritage"] == "deckhouse" &&
```
miklezzzz, line 167: Let's remove any mentioning of deckhouse product. As I understand there was an idea to add some extra env to the addon-operator like ADDON_OPERATOR_EXTRA_LABELS with default value `heritage=addon-operator`. This way it would be possible to redefine this env if needed to something like `heritage=not-addon-operator,foo=bar` and apply custom `heritage` value to modules` CRDs.

miklezzzz (commented):
some amends are required

pkg/addon-operator/operator.go
```diff
@@ -2136,6 +2163,16 @@ func (op *AddonOperator) CreateConvergeModulesTasks(state *module_manager.Module
 			doModuleStartup = true
 		}
 
+		// create ensureCRDsTasks
+		ensureCRDsTasks = append(ensureCRDsTasks, sh_task.NewTask(task.ModuleEnsureCRDs).
```
yalosev, line 2167: seems like we add ensure-crds task for every module. It could be a bit expensive, to scan filesystem every time for every module. 
I assume, we have to add a flag like `CRDsExist: bool` to the BasicModule and fill it somewhere in the loader:
1. https://github.com/flant/addon-operator/blob/main/pkg/module_manager/loader/fs/fs.go#L63
2. https://github.com/deckhouse/deckhouse/blob/main/deckhouse-controller/pkg/controller/loader.go#L106

@miklezzzz what do you think?

pkg/addon-operator/ensure_crds.go
juev: OK
but use `ADDON_OPERATOR_EXTRA_LABELS` variable

pkg/addon-operator/ensure_crds.go
juev: DONE

pkg/addon-operator/operator.go
miklezzzz: yep, makes sense, we would need:
1) add a flag to the basic module definition about having/not having CRDs to apply
2) set this flag somewhere around modulemanager's Init() call, I suppose module loader has to be updated to support CRD discovery

miklezzzz: Also, I personally think we should apply CRD's only on Startup, but I might be wrong..

miklezzzz: if someone is messing up with CRDs when operator is running it shouldn't be out problem, should it? @yalosev

yalosev: Does enabling a module generate an onStartup task?

juev: DONE

pkg/addon-operator/ensure_crds.go
juev: DONE

pkg/addon-operator/operator.go
miklezzzz: awesome, let's add something like ModuleHasCRDs method to the module manager and ask it before creating a moduleEnsureCRD task, it will result in having fewer tasks and less info when debugging/logging

miklezzzz (commented):
some rfc

pkg/addon-operator/operator.go
juev: DONE

juev: The only thing I think is that this excess overloads the logic. But it doesn't give much profit.

miklezzzz (approved):

pkg/module_manager/models/modules/basic.go
```diff
@@ -63,10 +65,13 @@ func NewBasicModule(name, path string, order uint32, staticValues utils.Values,
 		return nil, fmt.Errorf("new values storage: %w", err)
 	}
 
+	crdFilesPaths, _ := filepath.Glob(filepath.Join(path, "crds", "*.yaml"))
```
miklezzzz, line 68: could you please check that CRDs get isntalled even if they are located in subfolders, it seems some modules may have different structure of CRD directories like:
```
./crds/native/somecrd.yaml
```

pkg/app/app.go
```diff
@@ -146,6 +149,11 @@ func DefineStartCommandFlags(kpApp *kingpin.Application, cmd *kingpin.CmdClause)
 		Default(AppliedExtenders).
 		StringVar(&AppliedExtenders)
 
+	cmd.Flag("extra-labels", "String with CRDs label selectors, like `heritage=addon-operator`").
+		Envar("ADDON_OPERATOR_EXTRA_LABELS").
```
yalosev, line 153:
```suggestion
	cmd.Flag("crd-extra-labels", "String with CRDs label selectors, like `heritage=addon-operator`").
		Envar("ADDON_OPERATOR_CRD_EXTRA_LABELS").
```

pkg/addon-operator/ensure_crds.go
```diff
@@ -0,0 +1,222 @@
...
+	k8sTasks *multierror.Group
+
+	crdExtraLabels map[string]string
+}
+
+func (cp *CRDsInstaller) Run(ctx context.Context) *multierror.Error {
+	result := new(multierror.Error)
+
+	for _, crdFilePath := range cp.crdFilesPaths {
+		if match := strings.HasPrefix(filepath.Base(crdFilePath), "doc-"); match {
```
yalosev, line 69: I guess we can exclude these `doc-xxx` files on discovery phase

pkg/module_manager/models/modules/basic.go
yalosev, line 68: Let's exclude `doc-` files here and also exclude files started with `_`

juev: This function allows you to find files by mask only within one directory. If you need to use nested directories, you need to use filepath.Walk.

DONE

pkg/app/app.go
juev: DONE

pkg/addon-operator/ensure_crds.go
juev: DONE

pkg/module_manager/models/modules/basic.go
juev: DONE

juev:
After using Walk to find crd files, testing failed

I removed crd for multitenancy-manager and enabled it. Installation stopped with ClusterRoleBinding issue:

Screenshot 2024-08-31 at 20 26 57

docs/src/RUNNING.md
```diff
@@ -67,6 +67,10 @@ dev_cluster_live_ticks 32
 ...
 ```
 
+**ADDON_OPERATOR_CRD_EXTRA_LABELS** – string with CRDs label selectors. Default is `heritage=addon-operator`.
```
yalosev, line 70:
```suggestion
**ADDON_OPERATOR_CRD_EXTRA_LABELS** – string with CRDs label pairs.
For example: `heritage=my-app,scope=extra`
Default is `heritage=addon-operator`.
```

docs/src/RUNNING.md
```diff
@@ -67,6 +67,10 @@ dev_cluster_live_ticks 32
...
+
+**ADDON_OPERATOR_CRD_FILTERS** – String of filters for the CRD, separated by commas. Default is `doc-,_`.
```
yalosev, line 72: I guess it should be called somthing like `..._CRD_FILTER_PREFIXES`

pkg/addon-operator/operator.go
```diff
@@ -80,6 +85,10 @@ type AddonOperator struct {
 
 	// LeaderElector represents leaderelection client for HA mode
 	LeaderElector *leaderelection.LeaderElector
+
+	// CrdExtraLabels contains labels for processing CRD files
+	// like heritage=addon-operator
+	CrdExtraLabels map[string]string
```
yalosev, line 91:
```suggestion
	CRDExtraLabels map[string]string
```

like APIServices and so on

yalosev (approved):

PR #512, opened 2024-10-07
Metric storage mock

libmonsoon-dev:
#### Overview

Added an interface and corresponding mock to simplify testing of metrics related logic


miklezzzz (commented):
we usually accumulate changes before updating dependencies

yalosev:
Will bump in the another PR

PR #514, opened 2024-10-17
[helm_resources_manager] use cache instead of a plain client

miklezzzz:
#### Overview
Currently, helm resources manager creates a resources monitor for every enabled module that has a helm release. In its turn, a resource monitor periodically checks whether all the resource of the corresponding helm release are presented in the cluster. These checks are performed by retrieving partial metadata for every object of the release via List operation. The main gist of the PR is to get objects' metadata not from the cluster API directly, but rather from the cache, decreasing amount of List operations and speeding up some of the operations related to helm releases (checking if a helm release should be applied, running periodic checks for absent resources).

* Helm resources manager gets a cache and shares it with resource monitors it creates.
* The cache has labelSelector filter that should help cache only related objects (objects, deployed by the addon-operator).
* In order to tell the difference between related and not related objects, a post-renderer was introduced.
* The post-renderer applies an additional label (`heritage=addon-operator` by default, yet configurable) to every resource the operator deploys (see helm post-renderer functionality).
* The post-renderer is supported by both helm3 and helm3lib implementations (with quite noticeable discrepancies in realizations, so using helm3 is discouraged due to operational overhead).


pkg/helm_resources_manager/helm_resources_manager.go
```diff
@@ -44,25 +48,69 @@ type helmResourcesManager struct {
...
+		eventCh:    make(chan ReleaseStatusEvent),
+		monitors:   make(map[string]*ResourcesMonitor),
+		ctx:        cctx,
+		cancel:     cancel,
+		kubeClient: kclient,
+		cache:      cache,
+	}, nil
+}
+
+func (hm *helmResourcesManager) WithCache(ctx context.Context) error {
```
yalosev, line 88: This function is not used anywhere. Why do we need it?

miklezzzz: my bad, I forgot to delete it after merging WithClient and WithCache into NewHelmResourcesManager

pkg/helm/post_renderer/post_renderer.go
```diff
@@ -0,0 +1,39 @@
...
+		return renderedManifests, nil
+	}
+	nodes, err := kio.FromBytes(renderedManifests.Bytes())
+	if err != nil {
+		return nil, fmt.Errorf("parse rendered manifests failed: %v", err)
+	}
+	for _, v := range nodes {
+		labels := v.GetLabels()
+		if labels == nil {
+			labels = make(map[string]string)
```
yalosev, line 25:
```suggestion
			labels = make(map[string]string, len(p.ExtraLabels))
```

pkg/helm/post_renderer/post_renderer.go
```diff
@@ -0,0 +1,39 @@
...
+
+func (p *PostRenderer) Run(renderedManifests *bytes.Buffer) (modifiedManifests *bytes.Buffer, err error) {
+	if len(p.ExtraLabels) == 0 {
+		return renderedManifests, nil
+	}
+	nodes, err := kio.FromBytes(renderedManifests.Bytes())
+	if err != nil {
+		return nil, fmt.Errorf("parse rendered manifests failed: %v", err)
+	}
+	for _, v := range nodes {
```
yalosev, line 22:
```suggestion
	for _, node := range nodes {
```

pkg/helm/post_renderer/post_renderer.go
```diff
@@ -0,0 +1,39 @@
...
+		labels := v.GetLabels()
+		if labels == nil {
+			labels = make(map[string]string)
+		}
+		for k, v := range p.ExtraLabels {
+			labels[k] = v
+		}
+		v.SetLabels(labels)
```
yalosev, line 30:
```suggestion
		node.SetLabels(labels)
```

pkg/helm_resources_manager/test/mock/mock.go
```diff
@@ -17,6 +17,10 @@ func (h *MockHelmResourcesManager) WithContext(_ context.Context) {}
 
 func (h *MockHelmResourcesManager) WithKubeClient(_ *klient.Client) {}
 
+func (h *MockHelmResourcesManager) WithCache(_ context.Context) error {
+	return nil
+}
+
```
yalosev, line 23:
```suggestion

```

pkg/helm/post_renderer/post_renderer.go
```diff
@@ -0,0 +1,39 @@
...
+	if len(p.ExtraLabels) == 0 {
+		return renderedManifests, nil
+	}
+	nodes, err := kio.FromBytes(renderedManifests.Bytes())
+	if err != nil {
+		return nil, fmt.Errorf("parse rendered manifests failed: %v", err)
+	}
+	for _, node := range nodes {
+		labels := node.GetLabels()
+		if labels == nil {
```
ldmonster, line 24: тут желательно использовать
```go
if len(labels) == 0 {
...
}
```
т.к. len(nil) == 0, мы всегда получим однозначный результат

pkg/helm/version_detect.go
```diff
@@ -28,6 +31,11 @@ const (
 	Helm3Lib = ClientType("Helm3Lib")
 )
 
+func checkPostRenderer(path string) error {
+	_, err := os.Stat(path)
```
ldmonster, line 35: it's ok to return raw OS error to user here?

pkg/helm/version_detect_test.go
```diff
@@ -14,6 +14,8 @@ func Test_DetectHelmVersion(t *testing.T) {
 
 		_ = os.Setenv("HELM3", "yes")
 		defer os.Unsetenv("HELM3")
+		_ = os.Setenv("HELM_POST_RENDERER_PATH", "testdata/helm-fake/helm3/helm")
```
ldmonster, line 17: Maybe [this](https://pkg.go.dev/testing@master#T.Setenv) fits better here

cmd/addon-operator/main.go
```diff
@@ -151,29 +143,25 @@ func runHAMode(ctx context.Context, operator *addon_operator.AddonOperator) {
...
 	}
 
 	go func() {
 		<-ctx.Done()
 		log.Info("Context canceled received")
-		err := syscall.Kill(1, syscall.SIGUSR2)
-		if err != nil {
-			log.Infof("Couldn't shutdown addon-operator: %s\n", err)
-			os.Exit(1)
+		if err := syscall.Kill(1, syscall.SIGUSR2); err != nil {
```
ldmonster, line 163: We can use [notify context](https://pkg.go.dev/os/signal#NotifyContext) to catch syscalls
```go
ctx, stopNotify := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
defer stopNotify()
```

pkg/helm/post_renderer/post_renderer.go
```diff
@@ -0,0 +1,39 @@
...
+			labels = make(map[string]string, len(p.ExtraLabels))
+		}
+		for k, v := range p.ExtraLabels {
+			labels[k] = v
```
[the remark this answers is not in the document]
miklezzzz: yes, it goes along with the gist of the pr.

pkg/helm/version_detect.go
miklezzzz: yes, I don't care much about a deprecated way of deploying releases. All I want is to make sure  the post-renderer binary is in place.

pkg/helm/post_renderer/post_renderer.go
miklezzzz: explain, see no difference between nil and len

pkg/helm/version_detect_test.go
miklezzzz: It could be. But it would be rather matter of refactoring of  all tests.

pkg/helm/post_renderer/post_renderer.go
ldmonster: 1) it's more clear. no difference between 2 situations. if len(nil) == 0 - allocate map, if len(map) == 0 - allocate map. situations in which an error can occur becomes smaller.
2) if we have empty map for some reason - we can accurately allocate memory for it. idk how long labels map is. if we can allocate the required amount of memory at once, why not do this in both options?

miklezzzz: well, actually, I have to admit that this code is faulty as GetLabels() returns an empty map, which doesn't equal to nil, in case there aren't any labels in the node. I just copy-pasted it without considering for a second.

ldmonster (approved):

PR #516, opened 2024-10-23
[addon-operator] Replace Logrus with slog

ldmonster:
#### Overview

Replace Logrus to Slog implementation 

We can freely remove Logrus. That's the point!


pkg/module_manager/go_hook/go_hook.go
```diff
@@ -41,13 +44,43 @@ type FilterResult interface{}
...
+}
+
 type HookInput struct {
 	Snapshots        Snapshots
 	Values           *PatchableValues
 	ConfigValues     *PatchableValues
 	MetricsCollector MetricsCollector
 	PatchCollector   *object_patch.PatchCollector
-	LogEntry         *logrus.Entry
+	Logger           ILogger
```
juev, line 83: Why are we using such a big interface here?

cmd/addon-operator/main.go
```diff
@@ -143,25 +153,27 @@ func runHAMode(ctx context.Context, operator *addon_operator.AddonOperator) {
...
+		unilogger.Fatal("with leader election", slog.String("error", err.Error()))
 	}
 
 	go func() {
 		<-ctx.Done()
-		log.Info("Context canceled received")
+		unilogger.Info("Context canceled received")
 		if err := syscall.Kill(1, syscall.SIGUSR2); err != nil {
-			log.Fatalf("Couldn't shutdown addon-operator: %s\n", err)
+			log.Fatalf("Couldn't shutdown addon-operator: %s\n", err.Error())
```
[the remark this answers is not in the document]
ldmonster: because thats by bad! thank you! done

pkg/module_manager/go_hook/go_hook.go
ldmonster: because we need to refactor our hooks, before trim this interface.
we want to add this logger first, and refactor hooks in second iteration

juev: Using an interface (especially such a complex one) and only for refactoring, it seems that it is not necessary

yalosev: We can't migrate in one iteration, so it's a temporary interface. Only one thing we have to do here, it's to comment `XxxxF()` methods with line:
```
// Deprecated: use Debug instead
```

ldmonster: done

pkg/addon-operator/operator.go
```diff
@@ -381,7 +402,7 @@ func (op *AddonOperator) RegisterManagerEventsHandlers() {
 			"event.id": uuid.Must(uuid.NewV4()).String(),
 			"binding":  string(htypes.Schedule),
 		}
-		logEntry := log.WithFields(utils.LabelsToLogFields(logLabels))
+		logEntry := utils.EnrichLoggerWithLabels(op.Logger, logLabels)
```
juev, line 405: I wonder why we can't put the enrichment with lables in the log package? Why do we need to create separate functions in the common utils package?

pkg/kube_config_manager/backend/configmap/configmap.go
```diff
@@ -6,7 +6,7 @@ import (
 	"strings"
 	"time"
 
-	log "github.com/sirupsen/logrus"
+	log "github.com/deckhouse/deckhouse/pkg/log"
```
juev, line 9: Our package and so-called log, do not need to set alice for it

pkg/addon-operator/operator.go
ldmonster: we can add this to the next iteration of logger development.
greate discussion point

pkg/kube_config_manager/backend/configmap/configmap.go
ldmonster: done

yalosev (approved):

PR #528, opened 2024-11-12
Improved CRDs Installer

libmonsoon-dev:
#### Overview
Abstracted the kubernetes client for use in tests

### DIFF
--- a/pkg/addon-operator/ensure_crds.go
+++ b/pkg/addon-operator/ensure_crds.go
@@ -105,6 +105,46 @@ func (cp *CRDsInstaller) Run(ctx context.Context) *multierror.Error {
 	return result
 }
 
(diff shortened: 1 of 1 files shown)


pkg/addon-operator/ensure_crds.go
```diff
@@ -105,6 +105,46 @@ func (cp *CRDsInstaller) Run(ctx context.Context) *multierror.Error {
...
+			continue
+		}
+
+		for _, version := range crd.Spec.Versions {
+			gvr := schema.GroupVersionResource{
+				Group:    crd.Spec.Group,
+				Version:  version.Name,
+				Resource: crd.Spec.Names.Plural,
+			}
+			list, err := cp.k8sClient.Resource(gvr).List(ctx, apimachineryv1.ListOptions{})
```
yalosev, line 127: It's useless to check resources in every version. API-server will convert all existent resource to the desired version. The best way is to fetch version with `storage: true` and check resources in that version

PR #546, opened 2024-12-18
feat: replace deprecated logger methods

juev:
#### Overview

Fixed: #524

Just replace deprecated logger methods

nope

### DIFF
--- a/pkg/helm/helm3/helm3.go
+++ b/pkg/helm/helm3/helm3.go
@@ -3,6 +3,7 @@ package helm3
 import (
 	"bytes"
 	"fmt"
+	"log/slog"
 	"os/exec"
 	"sort"
 	"strings"
@@ -87,7 +88,7 @@ func (h *Helm3Client) initAndVersion() error {
 	}
 	stdout = strings.Join([]string{stdout, stderr}, "\n")
 	stdout = strings.ReplaceAll(stdout, "\n", " ")
-	log.Infof("Helm 3 version: %s", stdout)
+	log.Info("Helm 3 version", slog.String("version", stdout))
 
 	return nil
 }
@@ -161,12 +162,20 @@ func (h *Helm3Client) UpgradeRelease(releaseName string, chart string, valuesPat
 		args = append(args, setValue)
 	}
 
-	h.Logger.Infof("Running helm upgrade for release '%s' with chart '%s' in namespace '%s' ...", releaseName, chart, namespace)
+	h.Logger.Info("Running helm upgrade for release ...",
+		slog.String("release", releaseName),
+		slog.String("chart", chart),
+		slog.String("namespace", namespace))
 	stdout, stderr, err := h.cmd(args...)
 	if err != nil {
 		return fmt.Errorf("helm upgrade failed: %s:\n%s %s", err, stdout, stderr)
 	}
-	h.Logger.Infof("Helm upgrade for release '%s' with chart '%s' in namespace '%s' successful:\n%s\n%s", releaseName, chart, namespace, stdout, stderr)
+	h.Logger.Info("Helm upgrade for release successful",
+		slog.String("release", releaseName),
+		slog.String("chart", chart),
+		slog.String("namespace", namespace),
+		slog.String("stdout", stdout),
+		slog.String("stderr", stderr))
 
 	return nil
 }
@@ -191,7 +200,7 @@ func (h *Helm3Client) GetReleaseValues(releaseName string) (utils.Values, error)
 }
 
 func (h *Helm3Client) DeleteRelease(releaseName string) (err error) {
-	h.Logger.Debugf("helm release '%s': execute helm uninstall", releaseName)
+	h.Logger.Debug("helm release: execute helm uninstall", slog.String("release", releaseName))
 
 	args := []string{
 		"uninstall", releaseName,
@@ -202,7 +211,7 @@ func (h *Helm3Client) DeleteRelease(releaseName string) (err error) {
 		return fmt.Errorf("helm uninstall %s invocation error: %v\n%v %v", releaseName, err, stdout, stderr)
 	}
 
-	h.Logger.Debugf("helm release %s deleted", releaseName)
+	h.Logger.Debug("helm release deleted", slog.String("release", releaseName))
 	return
 }
 
@@ -230,9 +239,9 @@ func (h *Helm3Client) ListReleasesNames() ([]string, error) {
 		return nil, fmt.Errorf("helm list failed: %v\n%s %s", err, stdout, stderr)
 	}
 
-	list := []struct {
+	var list []struct {
 		Name string `json:"name"`
-	}{}
+	}
 
 	if err := k8syaml.Unmarshal([]byte(stdout), &list); err != nil {
 		return nil, fmt.Errorf("helm list returned invalid json: %v", err)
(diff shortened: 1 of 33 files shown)


pkg/module_manager/models/modules/global.go
```diff
@@ -194,6 +195,10 @@ func (gm *GlobalModule) executeHook(h *hooks.GlobalHook, bindingType sh_op_types
 		return fmt.Errorf("global hook '%s' failed: %s", h.GetName(), err)
 	}
 
+	if hookResult == nil {
+		return nil
+	}
+
```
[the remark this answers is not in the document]
juev: A rather strange logic is used here,
1. We calculate the result `hookResult, err := h.Execute(h.GetConfigVersion(), bc, "global", prefixedConfigValues, prefixedValues, logLabels)`
2. Next, we check that the result is not null and there is a Usage to display the result.
3. Only after that we check for the error, and if there is one, return it.
4. And after that, we are already trying to fully access the hookResult fields, although theoretically it is possible that we get nil. As a result, instead of a mistake, we will get panic.
5. To avoid panic, we do a check on nil before accessing the hookResult fields. Since we've already checked for errors and they're no longer there, since we've got a hookResult with nil, we'll just exit the function with the result nil.

In a good way, after the calculation, you need to immediately check the error, if there is one, return it. Then check for nil, and then decide what to do, return a new error or return nil. And then process the result.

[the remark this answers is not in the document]
miklezzzz: yeah, the logic is strange, or even tightly coupled with what h.Execute() returns, but if there is no possibility of getting nil without errors, does it make sense to check? :)

miklezzzz (approved):

pkg/module_manager/models/modules/global.go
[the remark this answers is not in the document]
miklezzzz: and if we check and get nil for the result, but w/o an error, it looks to me like an error condition :thinking: . I think we'd better return a custom error like 'something went wrong' instead of concealing the fact that we'd got nothing.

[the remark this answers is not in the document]
juev: It makes sense to think about it

pkg/helm/helm3/helm3.go
```diff
@@ -230,9 +239,9 @@ func (h *Helm3Client) ListReleasesNames() ([]string, error) {
 		return nil, fmt.Errorf("helm list failed: %v\n%s %s", err, stdout, stderr)
 	}
 
-	list := []struct {
+	var list []struct {
```
ldmonster, line 242: can you explain this please?

juev: Initializing a list to nil is preferable to using var rather than the assignment operator.

pkg/module_manager/models/modules/global.go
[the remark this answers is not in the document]
ldmonster: if we got an nil hookresult, we must get non nil error.
if we not getting non nil error and hookresult is nil - it's a bug on Execute() function.

do not need to double check for nil

pkg/module_manager/models/modules/basic.go
```diff
@@ -360,7 +361,8 @@ func RecursiveGetBatchHookExecutablePaths(dir string, logger *log.Logger) ([]str
 }
 
 var (
-	ErrFileHasNotMetRequirements   = errors.New("file has not met requirements")
+	// ErrFileHasNotMetRequirements   = errors.New("file has not met requirements")
```
ldmonster, line 364: delete if we not need it.

pkg/module_manager/models/modules/synchronization_state.go
```diff
@@ -61,14 +62,16 @@ func (s *SynchronizationState) IsCompleted() bool {
...
+				slog.String("hook", state.HookName),
+				slog.String("binding", state.BindingName))
 			break
 		}
 	}
 	return done
 }
 
-// func (s *SynchronizationState) QueuedForBinding(metadata task.HookMetadata) {
+// QueuedForBinding ...
```
ldmonster, line 74: needs to be done, or deleted

pkg/module_manager/module_manager.go
```diff
@@ -1233,12 +1248,12 @@ func queueHasPendingModuleDeleteTask(q *queue.TaskQueue, moduleName string) bool
 // registerModules load all available modules from modules directory.
 func (mm *ModuleManager) registerModules(scriptEnabledExtender *script_extender.Extender) error {
 	if mm.ModulesDir == "" {
-		log.Warnf("Empty modules directory is passed! No modules to load.")
+		log.Warn("Empty modules directory is passed! No modules to load.")
```
ldmonster, line 1251: interesting discussion point about points in logs

pkg/module_manager/loader/fs/fs.go
```diff
@@ -72,7 +74,7 @@ func (fl *FileSystemLoader) getBasicModule(definition moduleDefinition, commonSt
 	return m, nil
 }
 
-// reads single directory and returns BasicModule
+// LoadModule reads single directory and returns BasicModule
```
ldmonster, line 77: Prefferable to delete useless comment, not to refactor it.
Whats purpose of this comment?

It looks obvious.
LoadModule by given path and return BasicModule.

pkg/module_manager/models/modules/global.go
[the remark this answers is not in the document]
juev: And I think that double checking won’t hurt here. There may be bugs in the Execute function, and if they appear, there will be panic. Why do we need unnecessary panic at work? Easier to check.

pkg/module_manager/models/modules/basic.go
juev: done

pkg/module_manager/models/modules/synchronization_state.go
juev: done

pkg/module_manager/loader/fs/fs.go
juev: LoadModule is a public function, which means it must contain a docstring to generate documentation on pkg.go.dev

Another question is that the documentation should be more detailed. But at a minimum, its writing must meet the standards.

pkg/module_manager/models/modules/global.go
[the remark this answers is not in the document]
ldmonster: 1) Unnecessary panic notify us about this bag. With this realisation we skip critical bug.
2) This code change not often to make this check
3) golang has a error handling style. If i see correctly, we not to check anytime and everywhere for nil result from functions in our codebase. Can you link best practice about double checking?

pkg/module_manager/loader/fs/fs.go
ldmonster: 1) Addon operator do not used as a library from external developers. 
2) We do not need to cover clear code by obvious comments.

juev: This is where our opinions differ :)

pkg/module_manager/models/modules/global.go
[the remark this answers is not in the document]
ldmonster: If you want to guarantee that we'll never get a nil from Execute() function - you can cover every Execute function with tests in next pr.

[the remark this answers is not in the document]
juev: As you wish :)

pkg/module_manager/loader/fs/fs.go
ldmonster: I think not only mine is different. We have a lot of [codebase](https://github.com/deckhouse/dmt/blob/main/internal/flags/flags.go) without comment coverage.
Looks like we need to start apply practices, that we'll bring in discussions.

ldmonster (approved):

PR #560, opened 2025-02-05
chrooted bash executor

miklezzzz:
#### Overview
Implements running modules' shell scripts (hooks, enabled scripts) in a chrooted environment, prepared by a new module manager's component called environment manager.
The environment manager is enriched with the list of objects to provide in each module's chrooted directory in order to successfully launch corresponding module's scripts. Every object belongs to one of the predefined environment levels: EnabledScriptEnvironment and ShellHookEnvironment.
The resulting environment for a module is derived from the module's content (if the module has an enabled script or not, if the module has shell hooks to run or not, etc). The environment manager is capable of promoting and demoting modules' environments based on the aforementioned modules' content and the fact if a module is enabled or not.

Also, it contains several fixes to get rid of races when updating modules' states and module sets.


pkg/module_manager/models/hooks/kind/shellhook.go
```diff
@@ -176,6 +180,7 @@ func (sh *ShellHook) getConfig() ([]byte, error) {
 	envs := make([]string, 0)
 	envs = append(envs, os.Environ()...)
 	args := []string{"--config"}
+	var stderrBuf bytes.Buffer
```
miklezzzz, line 183: delete, as it was added for the sake of debugging

pkg/module_manager/environment_manager/evironment_manager.go
```diff
@@ -0,0 +1,225 @@
...
+}
+
+func (m *Manager) DisassembleEnvironmentForModule(moduleName, modulePath string, targetEnvironment Environment) error {
+	logEntry := utils.EnrichLoggerWithLabels(m.logger, map[string]string{
+		"operator.component": "EnvironmentManager.DisassembleEnvironmentForModule",
+	})
+	m.l.Lock()
+	defer m.l.Unlock()
+
+	currentEnvironment := m.preparedEnvironments[moduleName]
```
juev, line 82: What if we don't have a prepared environment for the module yet?

miklezzzz: it returns 0, default value for an Int variable, which, according to IOTA, equals to `NoEnvironment` :)

miklezzzz:
It seems I've found other races to fix before reviewing, though.

juev (approved):

ldmonster (approved):

pkg/module_manager/models/hooks/kind/shellhook.go
```diff
@@ -185,7 +188,8 @@ func (sh *ShellHook) getConfig() ([]byte, error) {
 		WithLogProxyHookJSON(shapp.LogProxyHookJSON).
 		WithLogProxyHookJSONKey(sh.LogProxyHookJSONKey).
 		WithLogger(sh.Logger.Named("executor")).
-		WithCMDStdout(nil)
+		WithCMDStdout(nil).
+		WithChroot(utils.GetModuleChrootPath(sh.moduleName))
```
yalosev, line 192: We are already have the `moduleSafeName` variable as input to this function. Could we use it for the chroot and not pass a new variable `moduleName` to the hook. It's not a huge mistake but we are leading again to some cycle dependencies: `module (with name)` -> `hook` -> `moduleName`. It's still better then pointer to a module inside a hook, but if we could use already presented variable (or mutate it if needed) - that would be nce

miklezzzz: I also found a couple of direct `exec.Command` calls in the module's context and converted them to executor calls, so that they would come along with the `chroot` conception. (Tests are provided, also).

yalosev (approved):

PR #566, opened 2025-03-04
[addon-operator] fix nil, add error handling, fix logger

ldmonster:
#### Overview

After we return nil helm module, we must not to use render (because of nil).
Added new error and handle it, where it needs

Inside touched files:
Add errors wrapping
Replace global logger with local logger


miklezzzz (approved):
a good catch, though I would do something with double error checking around RunHelmInstall, looks pretty ugly :(

yalosev (approved):

PR #579, opened 2025-03-21
Add module Unmanaged mode

miklezzzz:
#### Overview

- Adds an additional value key (similar to moduleName`Enabled`), named moduleName`ManagementState`, which can take only `Unmanaged` value so far. Putting a module into `Unmanaged` mode disables subsequent reconciles of the module's state in the cluster. All resources of the module get an additional label `managementState`: `Unmanaged` (if helm3 lib is in use) to make it clear that the module is running in `Unmanaged` mode.

- `addon-operator module render` command has been updated to support `--diff` flag that provides simple diff between manifests of the module in the corresponding helm release and the module's manifests deployed in the cluster.

- Helm factory has been updated to support setting additional labels conditionally.

- Helm kube client creator has been refactored to make it more straightforward. 

- Helm kube client has been updated to implement RestClientGetter interface, required for diffing modules` manifests.


pkg/addon-operator/debug_server.go
```diff
@@ -165,7 +179,114 @@ func (op *AddonOperator) RegisterDebugModuleRoutes(dbgSrv *debug.Server) {
...
+
+		f, err := os.CreateTemp("", "*")
+		if err != nil {
+			return nil, fmt.Errorf("failed to write helm chart manifests to a temporary directory: %w", err)
+		}
+
+		defer f.Close()
+
+		// file is deleted, yet the file descriptor isn't closed yet and can be accessed
+		err = os.Remove(f.Name())
```
yalosev, line 199: why not defer Remove() ?

miklezzzz: it's a "trick" :)
we create a temp file, open it for writing and delete it so the file isn't accessible from the file system, yet it is accessible from the "proc" file system as a file descriptor, so only the owner or a privileged user can read from it.
 
P.s. this is the way temp files syscall works in Linux.

yalosev (approved):

PR #583, opened 2025-04-01
feat: add GetReleaseChecksum method and update UpgradeRelease signature

juev:
#### Overview

Transferring the release checksum storage from variables to labels

Transfer the storage of checksums from environment variables to release tags

### DIFF
--- a/pkg/helm/helm3lib/helm3lib.go
+++ b/pkg/helm/helm3lib/helm3lib.go
@@ -193,6 +193,9 @@ func (h *LibClient) upgradeRelease(releaseName string, chartName string, valuesP
 	upg.MaxHistory = int(options.HistoryMax)
 	upg.Timeout = options.Timeout
 	// copy labels to avoid modifying the original map
+	if upg.Labels == nil {
+		upg.Labels = make(map[string]string)
+	}
 	maps.Copy(upg.Labels, labels)
 
 	chart, err := loader.Load(chartName)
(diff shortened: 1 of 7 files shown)


pkg/helm/helm3lib/helm3lib_test.go
```diff
@@ -38,7 +38,7 @@ func TestHelm3LibUpgradeDelete(t *testing.T) {
 
 	cl := initHelmClient(t)
 
-	err := cl.UpgradeRelease("test-release", "testdata/chart", nil, nil, cl.Namespace)
```
ldmonster, line 41: i think we need to start write tests for helm methods behaviour.

f.e. add this method for LibClient (not interface) and test with it 
```go
// ListReleases returns list of releases
func (h *LibClient) ListReleases() ([]*release.Release, error) {
	l := action.NewList(actionConfig)
	// list all releases regardless of their state
	l.StateMask = action.ListAll
	list, err := l.Run()
	if err != nil {
		return nil, fmt.Errorf("helm list failed: %s", err)
	}

	return list, nil
}
```

pkg/helm/helm3lib/helm3lib.go
```diff
@@ -199,6 +200,13 @@ func (h *LibClient) upgradeRelease(releaseName string, chartName string, valuesP
 	upg.SkipCRDs = true
 	upg.MaxHistory = int(options.HistoryMax)
 	upg.Timeout = options.Timeout
+	// copy labels to avoid modifying the original map
+	if upg.Labels == nil {
+		upg.Labels = make(map[string]string)
+	}
+	if labels != nil {
+		maps.Copy(upg.Labels, labels)
```
yalosev, line 208: didn't get the clue. why not to pass labels directly to the upgrade cli. Do the upgradeCli modify them?

yalosev: also have to add to the installCli below

ldmonster (approved):

PR #592, opened 2025-04-24
[feature] add cel validation for module values

ipaqsa:
#### Overview

It provides CEL validation for config values. 

Example:
```
type: object
properties:
  allowNamespacesWithoutProjects:
    type: boolean
    default: true
    description: |
      If this option is disabled - prohibit to create a namespace which is not a part of a Project.

x-deckhouse-validations:
  - expression: self.allowNamespacesWithoutProjects == true
    message: |
      The `allowNamespacesWithoutProjects` parameter should be enabled.
```


pkg/values/validation/schemas.go
```diff
@@ -130,6 +131,13 @@ func validateObject(dataObj interface{}, s *spec.Schema, rootName string) error
 		return fmt.Errorf("validated data object have to be utils.Values or map[string]interface{}, got %v instead", reflect.TypeOf(v))
 	}
 
+	// Validate values against x-deckhouse-validation rules.
+	if values, ok := dataObj.(map[string]interface{}); ok {
```
ldmonster, line 135: why not to use it on 127 line in switch statement?

ipaqsa: Why ?

yalosev (approved):

PR #594, opened 2025-04-25
feat: deployment backend prototype based on Nelm

ilya-lesikov:


pkg/helm/nelm/nelm.go
```diff
@@ -0,0 +1,254 @@
...
+		opts.KubeContext = *getter.Context
+	}
+
+	if opts.HelmDriver == "" {
+		opts.HelmDriver = os.Getenv("HELM_DRIVER")
+	}
+
+	// FIXME(ilya-lesikov): this function does A LOT. And with what addon-operator expects from us
+	// in regards to logging, all of this is kind of a mess now
+	opts.context = action.SetupLogging(context.TODO(), "info", "", "off")
```
ldmonster, line 42: What if we do not want to use logboek logger? Can we pass our logger to use in nelm?

pkg/helm/nelm/nelm.go
```diff
@@ -0,0 +1,254 @@
...
+	return strconv.FormatInt(int64(releaseGetResult.Release.Revision), 10), releaseGetResult.Release.Status.String(), nil
+}
+
+func (c *NelmClient) UpgradeRelease(releaseName string, chartName string, valuesPaths []string, setValues []string, labels map[string]string, namespace string) error {
+	c.logger.Info("Running nelm install for release",
+		slog.String("release", releaseName),
+		slog.String("chart", chartName),
+		slog.String("namespace", namespace))
+
+	// XXX: for some reason upgrade was called twice previously?
```
yalosev, line 121: It's because of singleton nature of the helm3lib client.
It initializes once and all Capabilties and APIResources support are filled during the initialization process. If we dynamically add some runtime feature, like [here](https://github.com/deckhouse/deckhouse/blob/901eaae4fb463d8c3977b18fdf4375ec635b867d/candi/control-plane-kubeadm/config.yaml.tpl#L90), this initialized client could not find them and fails with something like: `no matches for kind "ValidatingAdmissionPolicy" in version "admissionregistration.k8s.io/v1beta1"`

So we just catch this error, reinitialize the client and try once more time. If the second try fails - ok, it's a permanent error, we just return it

pkg/helm/nelm/nelm.go
```diff
@@ -0,0 +1,254 @@
...
+	// XXX: why we always deployed with upg.SkipCRDs = true?
```
yalosev, line 122: We have another flow. We discovery CRDs at the startup and [install](https://github.com/flant/addon-operator/blob/b7c97010128922e066fed35e2e53e91834a250de/pkg/addon-operator/operator.go#L1362) it for all enabled modules before module render. It allows to avoid races between modules which are based on some CRDs, like monitoring

pkg/helm/nelm/nelm.go
```diff
@@ -0,0 +1,254 @@
...
+		return "", fmt.Errorf("get nelm release %q: %w", releaseName, err)
+	}
+
+	if checksum, ok := releaseGetResult.Release.Annotations["moduleChecksum"]; ok {
+		return checksum, nil
+	}
+
+	// XXX: here was a fallback to get "_addonOperatorModuleChecksum" value from last release values
+	// instead. In Nelm via action.ReleaseGet we don't expose values of last release, I don't think
+	// they are relevant when we already expose manifests. Do we really need this?
```
yalosev, line 166: We can't ignore it. It's just a migration while we try to move from old values-injected method to release-metadata method. I guess we will finish this migration before the nelm engine will be enabled by default

pkg/helm/nelm/nelm.go
ilya-lesikov: Not at the moment. Maybe we implement in addon-operator very simple [NelmLogger](https://github.com/werf/nelm/blob/c4a5105c0c5e6ec24024f63560aac2196baf002c/internal/log/interface.go#L9-L31) and I'll expose [this global](https://github.com/werf/nelm/blob/c4a5105c0c5e6ec24024f63560aac2196baf002c/internal/log/default.go#L4) so we can override it in addon-operator?

ldmonster: Nice! Interface has interesting methods like ~Pop and ~Push. Can we read about them anywhere?

pkg/helm/nelm/nelm.go
ilya-lesikov: Going to expose Values and do the fallback then.

pkg/helm/nelm/nelm.go
```diff
@@ -0,0 +1,254 @@
...
+
+func (c *NelmClient) UpgradeRelease(releaseName string, chartName string, valuesPaths []string, setValues []string, labels map[string]string, namespace string) error {
+	c.logger.Info("Running nelm install for release",
+		slog.String("release", releaseName),
+		slog.String("chart", chartName),
+		slog.String("namespace", namespace))
+
+	// XXX: for some reason upgrade was called twice previously?
+	// XXX: why we always deployed with upg.SkipCRDs = true?
+	// XXX: Nelm does not care about pending status and will upgrade pending release just fine. Does it mean we are going to be fine without this: https://github.com/flant/addon-operator/blob/b7c97010128922e066fed35e2e53e91834a250de/pkg/helm/helm3lib/helm3lib.go#L293-L300
```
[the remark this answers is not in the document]
ilya-lesikov: No release rollbacks or deletions it is.

pkg/helm/nelm/nelm.go
ilya-lesikov: Will provide an option to not install CRDs in the Nelm action.

pkg/helm/nelm/nelm.go
ilya-lesikov: No need for this in Nelm

pkg/helm/nelm/nelm.go
ilya-lesikov: Discussed this. Added a NelmLogger that we gonna pass to Nelm. NelmLogger methods are not implemented yet, just provided a general structure.

pkg/helm/nelm/nelm.go
```diff
@@ -0,0 +1,278 @@
...
+		return checksum, nil
+	}
+
+	if recordedChecksum, hasKey := releaseGetResult.Values["_addonOperatorModuleChecksum"]; hasKey {
+		if recordedChecksumStr, ok := recordedChecksum.(string); ok {
+			return recordedChecksumStr, nil
+		}
+	}
+
+	return "", fmt.Errorf("moduleChecksum label not found in nelm release %q", releaseName)
```
yalosev, line 162:
```suggestion
	return "", fmt.Errorf("moduleChecksum annotation not found in nelm release %q", releaseName)
```

PR #620, opened 2025-06-09
[scheduler] add bootstrapped extender and separate critical/functional modules

ipaqsa:
#### Overview

It adds the bootstrapped extender to provide "critical area" separating functional modules from critical.  

It is related with [it](https://github.com/deckhouse/deckhouse/pull/13906)

it continues by [it](https://github.com/flant/addon-operator/pull/631)

the graph
456461795-1bf3bbe9-0bef-4125-a988-ce97c6eae37f

Modules can now be categorized as critical/functional(require bootstrapped cluster or not).

It brings two main changes:

1. All functional modules have the same weight - 999(set by default). It means they are loaded after the system modules by default and their tasks run in parallel except modules that depend on other functional modules.

2. Functional modules are disabled until bootstrapped done by the bootstrapped extender

At the first converge(installing) functional modules disabled because the cluster not bootstrapped.

After bootstrapping(there is a ready worker node) global values changed and it triggers reconverge but at this time functional modules will be enabled.

### DIFF
--- a/pkg/module_manager/module_manager.go
+++ b/pkg/module_manager/module_manager.go
@@ -604,6 +604,19 @@ func (mm *ModuleManager) GetModule(name string) *modules.BasicModule {
 	return mm.modules.Get(name)
 }
 
+func (mm *ModuleManager) GetCritical(moduleName string) bool {
+	module := mm.GetModule(moduleName)
+	if module == nil {
+		return false
+	}
+
+	return module.GetCritical()
+}
+
+func (mm *ModuleManager) GetFunctionalDependencies() map[string][]string {
+	return mm.moduleScheduler.GetFunctionalDependencies()
+}
+
 func (mm *ModuleManager) GetModuleNames() []string {
 	return mm.modules.NamesInOrder()
 }
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -53,7 +53,7 @@ var (
 	CRDsFilters = "doc-,_"
 
 	// NumberOfParallelQueues defines the number of precreated parallel queues for parallel execution
-	NumberOfParallelQueues   = 15
+	NumberOfParallelQueues   = 20
 	ParallelQueuePrefix      = "parallel_queue"
 	ParallelQueueNamePattern = ParallelQueuePrefix + "_%d"
 )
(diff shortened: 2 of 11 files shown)


pkg/module_manager/module_manager.go
```diff
@@ -354,7 +357,7 @@ func (mm *ModuleManager) Init(logger *log.Logger) error {
 	if err != nil {
 		return fmt.Errorf("couldn't create static extender: %w", err)
 	}
-	if err := mm.moduleScheduler.AddExtender(staticExtender); err != nil {
+	if err = mm.moduleScheduler.AddExtender(staticExtender); err != nil {
```
ldmonster, line 360: why did you replace `if err:=` with `if err=`?
revert pls

pkg/module_manager/module_manager_hooks.go
```diff
@@ -100,7 +100,7 @@ func (mm *ModuleManager) registerGlobalModule(globalValues utils.Values, configB
 
 	// applies a scheduler extender to follow which modules get enabled/disabled by dynamic patches
 	dynamicExtender := dynamic_extender.NewExtender()
-	if err := mm.moduleScheduler.AddExtender(dynamicExtender); err != nil {
+	if err = mm.moduleScheduler.AddExtender(dynamicExtender); err != nil {
```
ldmonster, line 103:
```suggestion
	if err := mm.moduleScheduler.AddExtender(dynamicExtender); err != nil {
```

pkg/module_manager/scheduler/node/mock/node_mock.go
```diff
@@ -6,6 +6,7 @@ import (
 )
 
 type MockModule struct {
+	System                bool
```
ldmonster, line 9: rename?
```suggestion
	Critical                bool
```

pkg/module_manager/module_manager.go
ipaqsa: why ?

pkg/module_manager/scheduler/node/mock/node_mock.go
ipaqsa: yes, thanks

ipaqsa: done

ldmonster (approved):

PR #743, opened 2026-03-26
[feature] Add module dop metrics

RottenRat:
#### Overview
Metrics about the version of modules (in labels) can be exported via flant-integration to DOP.

This information is needed to understand in which clusters which versions of modules have been rolled out, in order to monitor who has the old versions left.


ipaqsa (changes requested):
Redesign it

